PR #6, opened 2021-03-22
Keeping rf_we high for CSR instruction whenever the EX stage stalls

silabs-oysteink:
SEC equivalent with assumptions as stated in the commit message.


rtl/cv32e40x_id_stage.sv
```diff
@@ -830,10 +833,9 @@ module cv32e40x_id_stage import cv32e40x_pkg::*;
 
       end else if (id_ex_pipe_o.csr_access) begin
        //In the EX stage there was a CSR access, to avoid multiple
-       //writes to the RF, disable rf_we.
+       //writes to the RF, disable csr_access (cs_registers will keep it's rdata as it was when csr_access was 1'b1).
```
Silabs-ArjanB, line 836: Could you please add a design assertion checking this assumption?

rtl/cv32e40x_cs_registers.sv
```diff
@@ -516,6 +521,19 @@ module cv32e40x_cs_registers import cv32e40x_pkg::*;
     endcase
   end
 
+  // This instance keeps the last valid csr_rdata
+  cv32e40x_csr #(
```
Silabs-ArjanB, line 525: I would not use cv32e40x_csr here as it is a bit misleading. csr_rdata_q is not a CSR. Just add the required RTL directly here instead of re-using this module.

silabs-oysteink: Agreed, fixed in latest commit.

rtl/cv32e40x_id_stage.sv
silabs-oysteink: Assertion added to sva file.

Silabs-ArjanB (approved):

PR #18, opened 2021-03-25
Dbg help

silabs-oivind:
SEC clean vs 922cc7c


Silabs-ArjanB (changes requested):
I think the manifest (.flist) file also requires an update.
If it is easy you can also parameterize the nr of write ports

Silabs-ArjanB (approved):

PR #19, opened 2021-03-26
Implemented LSU bus errors, excluding controller updates.

silabs-oysteink:
SEC equivalent with 1e35b273a8364a3a283df80f8bed9e09b8857ffb (No functional change, only 'plumbing')


rtl/cv32e40x_load_store_unit.sv
```diff
@@ -438,6 +446,32 @@ module cv32e40x_load_store_unit import cv32e40x_pkg::*;
...
 
+  //////////////////////////////////////////////////////////////////////////////
+  // Handle bus errors
+  //////////////////////////////////////////////////////////////////////////////
+
+  // Propagate last trans_addr to WB stage (in case of bus_errors in WB this is needed for mtval)
+  // In case of a detected error, updates to data_addr_wb_o will be
+  // blocked by the controller until the NMI is taken.
+  // TODO:OK: A store following a load with bus_error
+    // will not have data dependency on loaded data (NMI should happen before this store)
```
Silabs-ArjanB, line 457: That is not really needed in my opinion. It is however important that such following store (or load) if it also receives a bus error, will not update data_addr_wb_o. The NMI can be taken later if needed.

rtl/cv32e40x_load_store_unit.sv
```diff
@@ -438,6 +446,32 @@ module cv32e40x_load_store_unit import cv32e40x_pkg::*;
...
+
+  // Folowing block is within the EX stage
+  always_ff @(posedge clk, negedge rst_n)
+  begin
+    if(rst_n == 1'b0) begin
+      data_addr_wb_o <= 32'h0;
+    end else begin
+      // Update for valid addresses if not blocked by controller
+      if(!(block_addr_wb_i || data_err_wb_o) && (trans_valid && trans_ready)) begin
```
Silabs-ArjanB, line 466: Why do you use 'data_err_wb_o' here? I would think that its value should already be included by the controller into the value of block_addr_wb_i,

rtl/cv32e40x_controller.sv
```diff
@@ -74,6 +74,10 @@ module cv32e40x_controller import cv32e40x_pkg::*;
   input  logic        data_we_ex_i,
   input  logic        data_misaligned_i,
 
+  input  logic        data_err_wb_i,              // LSU caused bus_error in WB stage
+  input  logic [31:0] data_addr_wb_i,             // Current LSU address in WB stage
+  output logic        block_addr_wb_o,            // To LSU to prevent data_addr_wb_i updates between error and taken NMI
```
Silabs-ArjanB, line 79: I understand why you named it like this, but I do think it is a bit misleading (as it makes it seem that this is an output of the WB stage. Can you rename it to block_data_addr_o? Same renaming in other files. Of course the new name is also a bit misleading as it doesn't block the propagation of data_addr into EX, just its propagation into WB.

Silabs-ArjanB (changes requested):
See comments

rtl/cv32e40x_load_store_unit.sv
silabs-oysteink: You're right, I'll remove it.

rtl/cv32e40x_controller.sv
silabs-oysteink: Renamed

rtl/cv32e40x_load_store_unit.sv
silabs-oysteink: A store with dependencies on a load with bus errors may use an undefined result as part of the address, and could be a potential security risk. I updated the comment to clarify this.

Silabs-ArjanB: Agreed

rtl/cv32e40x_core.sv
```diff
@@ -522,6 +532,11 @@ module cv32e40x_core import cv32e40x_pkg::*;
     .m_c_obi_data_if       ( m_c_obi_data_if    ),
     // ID/EX pipeline
     .id_ex_pipe_i          ( id_ex_pipe         ),
+
+    .data_addr_wb_o        ( data_addr_wb       ),
+    .data_err_wb_o         ( data_err_wb        ),
+
+    .block_addr_wb_i       ( block_addr_wb      ),
```
Silabs-ArjanB, line 539: This seems to be dangling now. Can this be changed into: .block_data_addr_i ( block_data_addr ),

Silabs-ArjanB (changes requested):
See comments

rtl/cv32e40x_core.sv
silabs-oysteink: Fixed

Silabs-ArjanB (approved):

PR #26, opened 2021-04-09
Make the bhv wrapper formal-friendly

silabs-robin:
1) In order to reach some PMA-related preconditions, it is necessary to have control over the PMA parameters to the core, and hence these were added to the wrapper so they can propagate from tb top level.

2) The interfaces used to pass payloads between pipeline stages, they have modports specifying i/o direction and cannot be implicitly connected (`.*`) to modports of incompatible type, and hence the assertion bindings were instead connected to the interface instances themselves (instead of to the portlist of a submodule).

I have tested this in formal and it both elaborates and runs without problem. I also tested it in core-v-verif and it works there too (as far as the self-checking can go without step-and-compare).

### DIFF
--- a/bhv/cv32e40x_wrapper.sv
+++ b/bhv/cv32e40x_wrapper.sv
@@ -105,6 +105,7 @@ module cv32e40x_wrapper
 
   bind cv32e40x_mult:            core_i.ex_stage_i.mult_i           cv32e40x_mult_sva         mult_sva         (.*);
   bind cv32e40x_if_stage:        core_i.if_stage_i                  cv32e40x_if_stage_sva     if_stage_sva     (
+    // The SVA's monitor modport can't connect to a master modport, so it is connected to the interface instance directly:
     .m_c_obi_instr_if(core_i.m_c_obi_instr_if),
     .*);
   bind cv32e40x_controller:      core_i.id_stage_i.controller_i     cv32e40x_controller_sva   controller_sva   (.*);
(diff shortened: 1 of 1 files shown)


bhv/cv32e40x_wrapper.sv
```diff
@@ -101,12 +104,16 @@ module cv32e40x_wrapper
   // RTL Assertions
 
   bind cv32e40x_mult:            core_i.ex_stage_i.mult_i           cv32e40x_mult_sva         mult_sva         (.*);
-  bind cv32e40x_if_stage:        core_i.if_stage_i                  cv32e40x_if_stage_sva     if_stage_sva     (.*);
+  bind cv32e40x_if_stage:        core_i.if_stage_i                  cv32e40x_if_stage_sva     if_stage_sva     (
+    .m_c_obi_instr_if(core_i.m_c_obi_instr_if),
+    .*);
```
Silabs-ArjanB, line 109: Would it work to replace .m_c_obi_instr_if(core_i.m_c_obi_instr_if), by .m_c_obi_instr_if(core_i.if_stage_i.m_c_obi_instr_if), ? (That would be the more logical connection as now you are assuming how the 'core' level connects to the 'if stage')

Same remark for connection to load-store unit below.

Silabs-ArjanB (changes requested):
See comments

Silabs-ArjanB (approved):

PR #27, opened 2021-04-13
Added propagation of exception related signals to WB stage.

silabs-oysteink:
They remain unused for now, but will be used when the controller is rewritten.

### DIFF
--- a/rtl/cv32e40x_id_stage.sv
+++ b/rtl/cv32e40x_id_stage.sv
@@ -838,6 +838,10 @@ module cv32e40x_id_stage import cv32e40x_pkg::*;
         //          and LSU it not in use
         id_ex_pipe_o.pc                     <= if_id_pipe_i.pc;
         id_ex_pipe_o.instr                  <= if_id_pipe_i.instr;
+        // Overwrite instruction word in case of compressed instruction
+        if (if_id_pipe_i.is_compressed) begin
+          id_ex_pipe_o.instr.bus_resp.rdata   <= {16'h0, if_id_pipe_i.compressed_instr};
+        end
         id_ex_pipe_o.illegal_insn           <= illegal_insn;
         id_ex_pipe_o.ebrk_insn              <= ebrk_insn;
         id_ex_pipe_o.wfi_insn               <= wfi_insn;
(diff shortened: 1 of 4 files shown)


rtl/cv32e40x_id_stage.sv
```diff
@@ -822,11 +831,20 @@ module cv32e40x_id_stage import cv32e40x_pkg::*;
...
-          id_ex_pipe_o.pc                   <= if_id_pipe_i.pc;
-        end
-
         id_ex_pipe_o.branch_in_ex           <= ctrl_transfer_insn == BRANCH_COND;
+
+        // Propagate signals needed for exception handling in WB
+        // TODO:OK: Clock gating of pc if no existing exceptions
+        //          and LSU it not in use
+        id_ex_pipe_o.pc                     <= if_id_pipe_i.pc;
+        id_ex_pipe_o.instr                  <= if_id_pipe_i.instr;
```
Silabs-ArjanB, line 840: Why is 'instr' needed? Is it for mtval update and if so, is there a need to also have access to the compressed instruction (now only the uncompressed or decompressed instruction is forwarded)?

rtl/include/cv32e40x_pkg.sv
```diff
@@ -914,6 +862,83 @@ typedef struct packed {
...
+  logic         data_req;
+  logic         data_we;
+  logic [1:0]   data_type;
+  logic         data_sign_ext;
+  logic [1:0]   data_reg_offset;
+  logic         data_misaligned;
+  logic [5:0]   data_atop;
+
+  // PC of last executed branch
+  logic [31:0]  pc;
```
Silabs-ArjanB, line 904: Is this comment accurate? Should this signal be moved in the 'exception handling group'?

rtl/cv32e40x_id_stage.sv
silabs-oysteink: It's for mtval. I added the compressed instruction aswell.

rtl/include/cv32e40x_pkg.sv
silabs-oysteink: Fixed

Silabs-ArjanB (approved):

PR #28, opened 2021-04-16
Moving controller and interrupt controller to toplevel

silabs-oysteink:
Controller is refactored with two submodules, one for the two FSMs and one for the stall and bypass related logic. Controller and interrupt controller moved out of ID stage and into the toplevel.

SEC equivalent with 9830aa9c7a1c5605ad47b81bbb6cc79aaddd5fa1


Silabs-ArjanB (changes requested):
How about the manifest file?
Where assertions ported as well (if at all required)?

Silabs-ArjanB (approved):

PR #29, opened 2021-04-19
move expected result to signals outside sva

silabs-robin:
The expected result_o of the multiplier SVAs are moved to their own signals.
This makes it much easier to inspect failures in waveforms.

### DIFF
--- a/sva/cv32e40x_mult_sva.sv
+++ b/sva/cv32e40x_mult_sva.sv
@@ -42,32 +42,36 @@ module cv32e40x_mult_sva
   ////////////////////////////////////////
 
   // Check result for MUL
+  logic [31:0] mul_result = $signed(op_a_i) * $signed(op_b_i);
   a_mul_result : // check multiplication result for MUL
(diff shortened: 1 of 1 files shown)


sva/cv32e40x_mult_sva.sv
```diff
@@ -42,32 +42,36 @@ module cv32e40x_mult_sva
   ////////////////////////////////////////
 
   // Check result for MUL
+  logic [31:0] mul_result = $signed(op_a_i) * $signed(op_b_i);
```
Silabs-ArjanB, line 45: Please split into separate declaration and assignment. Same remark for the other *result signals.

Silabs-ArjanB (changes requested):
See comment

sva/cv32e40x_mult_sva.sv
silabs-robin: I have now split declarations from assignments.

Silabs-ArjanB (approved):

PR #51, opened 2021-05-19
Removed writeback bypass for branches 

silabs-halfdan:
Removed writeback bypass for branches to achieve NRET=1 and avoid unnecessary RVFI updates before switching to the new controller.

This change allows modelling the pipeline as if all instructions retire in the writeback stage, which will be the case when the new controller updates are in.  


Silabs-ArjanB (approved):
This is a non-verified and potentially incorrect RTL update performed to allow faster progress on the RVFI implementation. The correctness of this update will be checked later (if still relevant after moving to the new pipeline controller), see https://github.com/openhwgroup/cv32e40x/issues/52

PR #72, opened 2021-06-10
NMI clarifications added

Silabs-ArjanB:


eroom1966:
could we please add that the NMI are edge-sensitive, to distinguish from irq_* which are level-sensitive
Then I think this is fully clarified

Silabs-ArjanB:
Hi @Silabs-ArjanB, I think using the word edge or level sensitive is a bit weird in the context of our NMIs as there are no corresponding NMI _pins_ on the core. I could describe the NMIs being triggered by certain 'load/store fault _events_', which I think then would convey the same meaning as what you might call edge sensitive.

eroom1966:
'load/store fault events', is fine
Thx

eroom1966:
the two issues are resolved with these changes
Thx

silabs-oysteink (approved):

PR #92, opened 2021-06-22
Rvfi fixes

silabs-halfdan:
Update modelling related to new meaning of valid signal in RTL,
Fixed issue related to modelling pc_wdata when killing if, fixed issue with debug.

### DIFF
--- a/bhv/cv32e40x_rvfi.sv
+++ b/bhv/cv32e40x_rvfi.sv
@@ -26,7 +26,7 @@ module cv32e40x_rvfi
    input logic                                rst_ni,
 
    //// ID probes ////
-   input logic [31:0]                         pc_if_i,
+   input logic [31:0]                         pc_id_i,
    input logic                                instr_id_valid_i,
    input logic [31:0]                         jump_target_id_i,
    // LSU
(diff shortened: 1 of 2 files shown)


bhv/cv32e40x_rvfi.sv
```diff
@@ -470,20 +476,22 @@ module cv32e40x_rvfi
 
       //// ID Stage ////
       if(instr_id_valid_i) begin
-        is_debug_entry_id   <= (instr_id_valid_i) ? is_debug_entry_if : is_debug_entry_if || is_debug_entry_id;
+        is_debug_entry_id   <= is_debug_entry_if;
         debug    [STAGE_ID] <= is_debug_entry_id;
-        pc_wdata [STAGE_ID] <= (pc_set_i && is_jump_id) ? jump_target_id_i : pc_if_i;
+        pc_wdata [STAGE_ID] <= (pc_set_i && is_jump_id) ? jump_target_id_i : pc_id_i + 4;
```
Silabs-ArjanB, line 481: Why does it have to be +4 (can't this be a compressed instruction)? Why is the value in pc_if_i not correct?

bhv/cv32e40x_rvfi.sv
```diff
@@ -421,7 +427,7 @@ module cv32e40x_rvfi
 
   assign wb_valid = wb_valid_i &&
```
Silabs-ArjanB, line 428: Can you add a todo here? All of this should be done in the RTL isntead of here (although we should not have treated the illegal instruction as invalid in the RTL either).

Silabs-ArjanB (approved):

PR #99, opened 2021-06-24
Suppressing *_insn_o from decoder when deassert_we is asserted.

silabs-oysteink:
On a PMA exception or bus error, the instruction word could
still contain a seemingly valid instruction, although it
had an associated exception.

Not SEC clean.

### DIFF
--- a/rtl/cv32e40x_decoder.sv
+++ b/rtl/cv32e40x_decoder.sv
@@ -193,8 +193,6 @@ module cv32e40x_decoder import cv32e40x_pkg::*;
   assign rf_we                          = decoder_ctrl_mux.rf_we;                           
   assign csr_en_o                       = decoder_ctrl_mux.csr_en;
   assign csr_op                         = decoder_ctrl_mux.csr_op;                          
-  assign mret_insn_o                    = decoder_ctrl_mux.mret_insn;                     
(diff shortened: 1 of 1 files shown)


rtl/cv32e40x_decoder.sv
```diff
@@ -217,6 +211,15 @@ module cv32e40x_decoder import cv32e40x_pkg::*;
   assign rf_we_o              = deassert_we_i ? 1'b0        : rf_we;
   assign ctrl_transfer_insn_o = deassert_we_i ? BRANCH_NONE : ctrl_transfer_insn;
 
+  // Suppress special instructions
```
Silabs-ArjanB, line 214: Why is illegal_insn_o not being suppressed?

Silabs-ArjanB (changes requested):

rtl/cv32e40x_decoder.sv
silabs-oysteink: It was forgotten. Fixed in the last push

Silabs-ArjanB (approved):

Silabs-ArjanB:
There seems to be a system issue here. ECA check remains pending. I performed a check using the 'ECA validation tool' and that states that there is a 'valid ECA on file' for this committer.

Issue #101, opened 2021-06-24
CSRs that are updated autonomously need to be reported with same timing as RTL would read such CSR


Silabs-ArjanB:
All counter (e.g. mcycle, cycle, cycleh, etc.) CSRs and the MIP CSR are updated autonomosly irrespective of instruction retirement. RVFI needs to report  the _rdata values such that they match the value that a CSR read instruction would see them (so basically these values need to propagate through the RVFI pipeline starting in EX).

Issue #100, opened 2021-06-24
RVFI interface: Add _n and _we for performance counters and mip as well


Silabs-ArjanB:
The RVFI interface connection to the core currently does not have _n and _we signals for the performance counters and mip. These CSRs are RW CSRs and should have these signals (even though there currently is not a single writeable bit in mip).

With respect to mcycle/minstret the RVFI spec says:

> Incrementing those counters should happen "between instructions", this means for example that an instruction that isn't a CSR write to mcycle should always have rvfi_csr_mcycle_rdata == rvfi_csr_mcycle_wdata.

This also implies that rdata and wdata can be different for CSR writes to such counters.

PR #105, opened 2021-06-24
Moved autonomously updated registers to EX stage

silabs-halfdan:
Moved RVFI register reads for autonomously updated registers to the EX stage in the pipe modelling. #101 
Added _n and _we  signals. #100 


Silabs-ArjanB (approved):
Looks good. Let's discuss later whether we should perform all reads such as you now implemented the 'auto' CSRs.

PR #110, opened 2021-06-28
Made mulh_a/mulh_b independenf of valid_i.

silabs-oysteink:
Not SEC clean, fails on cutpoint on multiplier results when
being killed.

Reduces area.

### DIFF
--- a/rtl/cv32e40x_mult.sv
+++ b/rtl/cv32e40x_mult.sv
@@ -139,11 +139,11 @@ module cv32e40x_mult import cv32e40x_pkg::*;
           ready_o         = 1'b1;
           valid_o         = 1'b0;
         end else begin
-          mulh_a           = mulh_al;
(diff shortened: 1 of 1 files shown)


silabs-halfdan (approved):
Looks good!

Silabs-ArjanB (commented):
Could you try out an alternative w.r.t area that assumes valid_i = 1 inside the case statement and then just does the following after the case statement:

        if (!valid_i) begin
          mulh_state_next = MUL_ALBL;
          ready_o = 1'b1;
          valid_o = 1'b0;
          mulh_acc_next = '0;
  end

(In the current code it is not immediately obvious that mulh_acc_next = 0 will happen on kill)

(So valid_i would not be used at all inside the case statement)

silabs-oysteink:
I will try that, getting back with numbers

Silabs-ArjanB (approved):

PR #114, opened 2021-06-29
Jumps and branches will only be active for one cycle.

silabs-oysteink:
This will allow IF (and ID) to fill up with the jump/branch
target while EX stage may be stalled. Slightly improves performance.

Bugfix to debug entry killing of stages.

Requires an update to RVFI to properly detect jumps in ID (and possibly branches in EX)

### DIFF
--- a/rtl/cv32e40x_controller_fsm.sv
+++ b/rtl/cv32e40x_controller_fsm.sv
@@ -107,10 +107,15 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
 
   
   // Events in ID
+  logic jump_in_id;
   logic jump_taken_id;
 
   // Events in EX
+  logic branch_in_ex;
   logic branch_taken_ex;
+
+  logic branch_taken_n;
+  logic branch_taken_q;
   
   // Events in WB
   logic exception_in_wb;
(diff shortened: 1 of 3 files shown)


rtl/cv32e40x_controller_fsm.sv
```diff
@@ -171,8 +183,8 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
                               ex_wb_pipe_i.ecall_insn                 ? EXC_CAUSE_ECALL_MMODE     :
                               EXC_CAUSE_BREAKPOINT;
 
-  // wfi in wb
-  assign wfi_in_wb = ex_wb_pipe_i.wfi_insn && ex_wb_pipe_i.instr_valid;
+  // wfi in wb, if debug_mode we treat it as a NOP
```
Silabs-ArjanB, line 186: There seems to be similar logic in cv32e40x_i_decoder (line 326); why do we look at debug in two places? The decoder code might not be correct; shouldn't it suppress wfi_insn there?

Silabs-ArjanB (commented):
Can you comment on where WFI should be suppressed to NOP? Decode or controller?

Silabs-ArjanB (approved):

PR #116, opened 2021-06-30
Changed minstret to (hopefully) correctly count retired instructions.

silabs-oysteink:
Halting EX stage in case of minstret/h read in EX and any valid instruction in WB.
Halting to get correct value for reading minstret to GPR.

No errors for CSR minstret checks in ci_check.

Not SEC clean

### DIFF
--- a/rtl/cv32e40x_controller_fsm.sv
+++ b/rtl/cv32e40x_controller_fsm.sv
@@ -243,7 +243,7 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
                               !id_ex_pipe_i.lsu_misaligned)) && !debug_mode_q;
                                
   // Performance counter events
-  assign ctrl_fsm_o.mhpmevent.minstret = ex_wb_pipe_i.instr_valid; // todo:low Not correct; just put something here such that minstret counter will not get optimized away (and did not use wb_valid to avoid long timing path)
+  assign ctrl_fsm_o.mhpmevent.minstret = ex_wb_pipe_i.instr_valid && !exception_in_wb && !ctrl_fsm_o.kill_wb && !ctrl_fsm_o.halt_wb; // todo: Should maybe use wb_stage local instr_valid, as the current code remakes that here.
   assign ctrl_fsm_o.mhpmevent.load = 1'b0; // todo:low
   assign ctrl_fsm_o.mhpmevent.store = 1'b0; // todo:low
   assign ctrl_fsm_o.mhpmevent.jump = 1'b0; // todo:low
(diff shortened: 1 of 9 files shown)


rtl/cv32e40x_ex_stage.sv
```diff
@@ -340,6 +340,6 @@ module cv32e40x_ex_stage import cv32e40x_pkg::*;
                        (id_ex_pipe_i.div_en && div_valid) ||
                        (id_ex_pipe_i.csr_en && csr_valid) ||
                        previous_exception // todo:ab:remove
-                      ) && instr_valid;
+                      ) && instr_valid && !ctrl_fsm_i.halt_ex;
```
Silabs-ArjanB, line 343: Make ID and EX stage consistent. In ID the halt signal is factored into instr_valid.
This relates to the todo on line 103 in the original EX stage module. That todo needs to be understood first before this PR gets fixed/merged

rtl/include/cv32e40x_pkg.sv
```diff
@@ -1019,6 +1019,7 @@ typedef struct packed {
   logic        load_stall;            // Stall due to load operation
   logic        csr_stall;
   logic        wfi_stall;
+  logic        minstret_stall;        // Stall due to minstret/h read in EX
```
Silabs-ArjanB, line 1022: Can't this just be part of csr_stall? I see now that csr_stall still stall ID whereas this new stall stalls EX. In the future, when both stall EX, they should get merged, so you can ignore this remark for now.

rtl/cv32e40x_core.sv
```diff
@@ -602,6 +603,8 @@ module cv32e40x_core import cv32e40x_pkg::*;
     // From controller FSM
     .ctrl_fsm_i                 ( ctrl_fsm               ),
 
+    .csr_raddr_o                ( csr_raddr_ex           ),
```
Silabs-ArjanB, line 606: The CSR module outputs use inconsistent names; we can clean that up later (don't know a good naming strategy for this yet; can you just add a todo?)

rtl/cv32e40x_controller_fsm.sv
```diff
@@ -243,7 +243,7 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
                               !id_ex_pipe_i.lsu_misaligned)) && !debug_mode_q;
                                
   // Performance counter events
-  assign ctrl_fsm_o.mhpmevent.minstret = ex_wb_pipe_i.instr_valid; // todo:low Not correct; just put something here such that minstret counter will not get optimized away (and did not use wb_valid to avoid long timing path)
+  assign ctrl_fsm_o.mhpmevent.minstret = ex_wb_pipe_i.instr_valid && !exception_in_wb && !ctrl_fsm_o.kill_wb && !ctrl_fsm_o.halt_wb; // todo: Should maybe use wb_stage local instr_valid, as the current code remakes that here.
```
Silabs-ArjanB, line 246: Yes, please use local instr_valid and then exclude instructions which should never count as retired (ecall, etc.)

silabs-halfdan (approved):
Conceptually this means defining the minstret update as a side-effect of the instruction rather than an update happening between instructions.
I would say this is ok if clearly specified/documented.

Silabs-ArjanB (changes requested):

rtl/cv32e40x_ex_stage.sv
silabs-oysteink: Ah, yes I will make them consistent

rtl/include/cv32e40x_pkg.sv
silabs-oysteink: Agreed. The csr_stall currently does not check csr_addr, for ID stage that is still a part of operand_b, so it felt better to use EX stage for this as that actually has a concept of csr_addr.

silabs-oysteink:
Closing PR, as the current implementation is very fragile  and requires some more updates.

PR #117, opened 2021-06-30
Pma assertions

silabs-robin:
This PR adds assertions for the PMA.

It adds a IS_INSTR_SIDE parameter to the sva module and also a few signals for connecting to the obi bus.
Then it adds the assertions themselves (according to the PMA vPlan) plus some helper logic.
Covers were added to make sure formal isn't overconstrained, to guide simulation testing, and for the insight they provide.

_NB. I did not get anyone else to review the assertions (neither the assertions alone or compared to the vPlan)._

_NB. Formal fails because all coverage is not reachable from all pma configs. I did not figure out a solution to this._

### DIFF
--- a/bhv/cv32e40x_wrapper.sv
+++ b/bhv/cv32e40x_wrapper.sv
@@ -223,7 +223,8 @@ bind cv32e40x_sleep_unit:
     cv32e40x_mpu_sva
       #(.PMA_NUM_REGIONS(PMA_NUM_REGIONS),
         .PMA_CFG(PMA_CFG))
-  mpu_if_sva(.*);
+  mpu_if_sva(.pma_addr(pma_i.trans_addr_i),
+             .*);
 
   // TODO: Reintroduce once LSU PMA support has been properly implemented in the controller
   /*
--- a/bhv/cv32e40x_wrapper.sv
+++ b/bhv/cv32e40x_wrapper.sv
@@ -221,24 +221,37 @@ bind cv32e40x_sleep_unit:
     core_i.if_stage_i.mpu_i 
     cv32e40x_mpu_sva
       #(.PMA_NUM_REGIONS(PMA_NUM_REGIONS),
-        .PMA_CFG(PMA_CFG))
+        .PMA_CFG(PMA_CFG),
+        .IS_INSTR_SIDE(1))
   mpu_if_sva(.pma_addr(pma_i.trans_addr_i),
-             .pma_cfg(pma_i.pma_cfg),
+             .pma_cfg (pma_i.pma_cfg),
              .instr_memtype_o(core_i.instr_memtype_o),
-             .instr_addr_o(core_i.instr_addr_o),
-             .instr_req_o(core_i.instr_req_o),
-             .instr_gnt_i(core_i.instr_gnt_i),
+             .instr_addr_o   (core_i.instr_addr_o),
+             .instr_req_o    (core_i.instr_req_o),
+             .instr_gnt_i    (core_i.instr_gnt_i),
+             .data_memtype_o('X),
+             .data_addr_o   ('X),
+             .data_req_o    ('X),
+             .data_gnt_i    ('X),
              .*);
 
-  // TODO:low Reintroduce once LSU PMA support has been properly implemented in the controller
-  /*
   bind cv32e40x_mpu:
     core_i.load_store_unit_i.mpu_i
     cv32e40x_mpu_sva
       #(.PMA_NUM_REGIONS(PMA_NUM_REGIONS),
-        .PMA_CFG(PMA_CFG))
-  mpu_lsu_sva(.*);
-  */
+        .PMA_CFG(PMA_CFG),
+        .IS_INSTR_SIDE(0))
+  mpu_lsu_sva(.pma_addr(pma_i.trans_addr_i),
+             .pma_cfg (pma_i.pma_cfg),
+             .instr_memtype_o('X),
+             .instr_addr_o   ('X),
+             .instr_req_o    ('X),
+             .instr_gnt_i    ('X),
+             .data_memtype_o(core_i.data_memtype_o),
+             .data_addr_o   (core_i.data_addr_o),
+             .data_req_o    (core_i.data_req_o),
+             .data_gnt_i    (core_i.data_gnt_i),
+             .*);
 `endif //  `ifndef COREV_ASSERT_OFF
   
     cv32e40x_core_log
--- a/bhv/cv32e40x_wrapper.sv
+++ b/bhv/cv32e40x_wrapper.sv
@@ -226,6 +226,8 @@ bind cv32e40x_sleep_unit:
              .pma_cfg(pma_i.pma_cfg),
              .instr_memtype_o(core_i.instr_memtype_o),
              .instr_addr_o(core_i.instr_addr_o),
+             .instr_req_o(core_i.instr_req_o),
+             .instr_gnt_i(core_i.instr_gnt_i),
              .*);
 
   // TODO:low Reintroduce once LSU PMA support has been properly implemented in the controller
--- a/bhv/cv32e40x_wrapper.sv
+++ b/bhv/cv32e40x_wrapper.sv
@@ -225,6 +225,7 @@ bind cv32e40x_sleep_unit:
   mpu_if_sva(.pma_addr(pma_i.trans_addr_i),
              .pma_cfg(pma_i.pma_cfg),
              .instr_memtype_o(core_i.instr_memtype_o),
+             .instr_addr_o(core_i.instr_addr_o),
              .*);
 
   // TODO:low Reintroduce once LSU PMA support has been properly implemented in the controller
--- a/bhv/cv32e40x_wrapper.sv
+++ b/bhv/cv32e40x_wrapper.sv
@@ -224,6 +224,7 @@ bind cv32e40x_sleep_unit:
       #(.PMA_NUM_REGIONS(PMA_NUM_REGIONS),
         .PMA_CFG(PMA_CFG))
   mpu_if_sva(.pma_addr(pma_i.trans_addr_i),
+             .pma_cfg(pma_i.pma_cfg),
              .*);
 
   // TODO: Reintroduce once LSU PMA support has been properly implemented in the controller
--- a/bhv/cv32e40x_wrapper.sv
+++ b/bhv/cv32e40x_wrapper.sv
@@ -225,14 +225,10 @@ bind cv32e40x_sleep_unit:
         .IS_INSTR_SIDE(1))
   mpu_if_sva(.pma_addr(pma_i.trans_addr_i),
              .pma_cfg (pma_i.pma_cfg),
-             .instr_memtype_o(core_i.instr_memtype_o),
-             .instr_addr_o   (core_i.instr_addr_o),
-             .instr_req_o    (core_i.instr_req_o),
-             .instr_gnt_i    (core_i.instr_gnt_i),
-             .data_memtype_o('X),
-             .data_addr_o   ('X),
-             .data_req_o    ('X),
-             .data_gnt_i    ('X),
+             .obi_memtype(core_i.instr_memtype_o),
+             .obi_addr   (core_i.instr_addr_o),
+             .obi_req    (core_i.instr_req_o),
+             .obi_gnt    (core_i.instr_gnt_i),
              .*);
 
   bind cv32e40x_mpu:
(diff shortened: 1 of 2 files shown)


silabs-robin:
I can't see that I have permission to select reviewers, but it would be nice with some feedback from @Silabs-ArjanB and @silabs-hfegran .

sva/cv32e40x_mpu_sva.sv
```diff
@@ -52,10 +71,77 @@ module cv32e40x_mpu_sva import cv32e40x_pkg::*; import uvm_pkg::*;
...
+
+    // Find lowest region matching addr
+    for (int i = 0; i < PMA_NUM_REGIONS; i++) begin
+      if (({PMA_CFG[i].word_addr_low, 2'b00} <= pma_addr) && (pma_addr < {PMA_CFG[i].word_addr_high, 2'b00})) begin
+        pma_lowest_match = i;
+        break;
+      end
+    end
+  end
+  cov_pma_matchnone : cover property (@(posedge clk) (!is_pma_matched));
```
Silabs-ArjanB, line 137: For these cover properties (and in general all cover properties): If the cover point requires specific parameter settings in order to be met, then I prefer if we exclude them from formal verification and only include them in simulation. We used to use `ifndef FORMAL for this; please check with @strichmo how we should treat such coverpoints. In my opinion it is not worth the extra work to start waiving/reviewing non-covered coverage points reported by formal for a specific configuration.

sva/cv32e40x_mpu_sva.sv
```diff
@@ -69,9 +155,89 @@ module cv32e40x_mpu_sva import cv32e40x_pkg::*; import uvm_pkg::*;
...
+    assert property (@(posedge clk)
+                     is_pma_matched |-> ((0 <= pma_match_num) && (pma_match_num <= 16)))
+      else `uvm_error("mpu", "illegal cfg index")
+
+  // Bufferable
+  a_pma_obi_bufrequired :
+    assert property (@(posedge clk) bus_trans_bufferable
+                     |-> obi_memtype[0] ^ (!obi_memtype[0] && was_obi_waiting && !$past(obi_memtype[0])))
+      else `uvm_error("mpu", "obi should have had bufferable flag")
+  a_pma_obi_bufallowed :
```
Silabs-ArjanB, line 184: This 2nd assertion is a bit counter intuitive (but not wrong). If a_pma_obi_bufrequired is of the form A -> B then I think it would be more readable (and provide coverage for free) if this second assertion is written in the form !A -> !B.

sva/cv32e40x_mpu_sva.sv
```diff
@@ -69,9 +155,89 @@ module cv32e40x_mpu_sva import cv32e40x_pkg::*; import uvm_pkg::*;
...
+      else `uvm_error("mpu", "PMA region match wasn't lowest")
+  a_pma_match_index :
+    assert property (@(posedge clk)
+                     is_pma_matched |-> ((0 <= pma_match_num) && (pma_match_num <= 16)))
+      else `uvm_error("mpu", "illegal cfg index")
+
+  // Bufferable
+  a_pma_obi_bufrequired :
+    assert property (@(posedge clk) bus_trans_bufferable
+                     |-> obi_memtype[0] ^ (!obi_memtype[0] && was_obi_waiting && !$past(obi_memtype[0])))
```
Silabs-ArjanB, line 182: The '^ excuse condition' should not be needed for the data side PMA instance. Please exclude it there.

Also add comments on what is being done here as this code, although intelligent, is not easily readable.

sva/cv32e40x_mpu_sva.sv
```diff
@@ -69,9 +155,89 @@ module cv32e40x_mpu_sva import cv32e40x_pkg::*; import uvm_pkg::*;
...
+    assert property (@(posedge clk) obi_memtype[0]
+                     |-> bus_trans_bufferable ^ (!bus_trans_bufferable && was_obi_waiting && $past(obi_memtype[0])))
+      else `uvm_error("mpu", "obi should not have had bufferable flag")
+
+  // Cacheable
+  a_pma_obi_cacherequired :
+    assert property (@(posedge clk) bus_trans_cacheable
+                     |-> obi_memtype[1] ^ (!obi_memtype[1] && was_obi_waiting && !$past(obi_memtype[1])))
+      else `uvm_error("mpu", "obi should have had cacheable flag")
+  a_pma_obi_cacheallowed :
```
Silabs-ArjanB, line 194: Same as above: Write as !A -> !B

sva/cv32e40x_mpu_sva.sv
```diff
@@ -69,9 +155,89 @@ module cv32e40x_mpu_sva import cv32e40x_pkg::*; import uvm_pkg::*;
...
+    assert property (@(posedge clk) obi_memtype[1]
+                     |-> bus_trans_cacheable ^ (!bus_trans_cacheable && was_obi_waiting && $past(obi_memtype[1])))
+      else `uvm_error("mpu", "obi should not have had cacheable flag")
+
+  // OBI req vs PMA err
+  a_pma_obi_reqallowed :
```
Silabs-ArjanB, line 200: Same as above. In this case the antecedent should use the pma_err signal (and obi_req should be in the consequence part)

Silabs-ArjanB (changes requested):
Please have a look at my comments; I will re-assign to Øivind for further review

sva/cv32e40x_mpu_sva.sv
```diff
@@ -69,9 +155,89 @@ module cv32e40x_mpu_sva import cv32e40x_pkg::*; import uvm_pkg::*;
...
+    assert property (@(posedge clk)
+                     is_pma_matched |-> (pma_match_num == pma_lowest_match))
+      else `uvm_error("mpu", "PMA region match wasn't lowest")
+  a_pma_match_index :
+    assert property (@(posedge clk)
+                     is_pma_matched |-> ((0 <= pma_match_num) && (pma_match_num <= 16)))
+      else `uvm_error("mpu", "illegal cfg index")
+
+  // Bufferable
+  a_pma_obi_bufrequired :
```
silabs-oivind, line 180: The bufferable flag in PMA_CFG is not being checked here. We're relying on the RTL to set buf_trans_bufferable correctly. The assertions should make it's own expected version rather than relying on the RTL. The same goes for bus_trans_cacheable and pma_err.

silabs-oivind (changes requested):
See comment in cv32e40x_mpu_sva.

silabs-hfegran (commented):
I agree with the comments of @Silabs-ArjanB and @silabs-oivind. In general it looks good, perhaps a short description of what's being tested in the comment would be useful? (in most cases it's straight forward to deduce from the assertion, but it's generally more readable to see this in plain text and to compare with requirements in the verification plan)

sva/cv32e40x_mpu_sva.sv
```diff
@@ -69,9 +155,89 @@ module cv32e40x_mpu_sva import cv32e40x_pkg::*; import uvm_pkg::*;
         assert (!PMA_CFG[i].cacheable) else `uvm_error("mpu", "PMA regions configured as I/O cannot be defined as cacheable")
       end
     end
-    
   end
-  
+
+  a_pma_valid_num_regions :
+    assert property (@(posedge clk)
```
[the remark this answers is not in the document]
silabs-robin: This is indeed nice for consistency. But I am not 100% sure that it is the best choice.
I have fixed this now.

We have a some other SVA files with assertions that don't use "disable iff":
https://github.com/openhwgroup/cv32e40x/blob/master/sva/cv32e40x_id_stage_sva.sv
https://github.com/openhwgroup/cv32e40x/blob/master/sva/cv32e40x_controller_fsm_sva.sv
https://github.com/openhwgroup/cv32e40x/blob/master/sva/cv32e40x_if_stage_sva.sv
https://github.com/openhwgroup/cv32e40x/blob/master/sva/cv32e40x_load_store_unit_sva.sv
~~https://github.com/openhwgroup/cv32e40x/blob/master/sva/cv32e40x_mpu_sva.sv~~
https://github.com/openhwgroup/cv32e40x/blob/master/sva/cv32e40x_prefetch_unit_sva.sv

bhv/cv32e40x_wrapper.sv
```diff
@@ -221,18 +221,37 @@ bind cv32e40x_sleep_unit:
...
     cv32e40x_mpu_sva
       #(.PMA_NUM_REGIONS(PMA_NUM_REGIONS),
-        .PMA_CFG(PMA_CFG))
-  mpu_lsu_sva(.*);
-  */
+        .PMA_CFG(PMA_CFG),
+        .IS_INSTR_SIDE(0))
+  mpu_lsu_sva(.pma_addr(pma_i.trans_addr_i),
+             .pma_cfg (pma_i.pma_cfg),
+             .instr_memtype_o('X),
```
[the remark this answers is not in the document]
silabs-robin: Done.

sva/cv32e40x_mpu_sva.sv
```diff
@@ -52,10 +71,77 @@ module cv32e40x_mpu_sva import cv32e40x_pkg::*; import uvm_pkg::*;
...
+
+
+  // Assign local signals depending on instr/data side instantiation
+
+  logic [ 1:0] obi_memtype;
+  logic [31:0] obi_addr;
+  logic        obi_req;
+  logic        obi_gnt;
+  generate
+    if (IS_INSTR_SIDE) begin
```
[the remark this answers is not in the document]
silabs-robin: Done.

sva/cv32e40x_mpu_sva.sv
silabs-robin: Done.

I have used ifndef to exclude the covers from formal.
If you @strichmo have any good ideas I can implement those later.
FYI, the problem here is that it is tricky and likely laborious to write generic conditions that enable/disable covers according to whatever pma config is currently given.

sva/cv32e40x_mpu_sva.sv
silabs-robin: Done.

sva/cv32e40x_mpu_sva.sv
silabs-robin: Done.

sva/cv32e40x_mpu_sva.sv
silabs-robin: I don't know if I agree.
I do think that writing `a <-> b` as `a -> b` and `!a -> !b` gives a free related cover, but this case is different.

Consider: `!req` does not imply `err`.

_(unless if I am thinking wrong right now)_

sva/cv32e40x_mpu_sva.sv
silabs-robin: Done.

> The assertions should make it's own expected version rather than relying on the RTL

I think I get the point, but there are 2 separate ideas that perhaps must be sorted out...
1) These assertions (buf, cache, err) are for checking the OBI bus signals. For that end, it doesn't matter too much if the RTL or expectation is used, I think.
2) The RTL must match expectations. I am pondering A) whether rtl-vs-expectations should be asserted separately from obi, and B) whether rtl-vs-expectations already are checked via the assertions regarding indices/matching. I think, yes it should be separate from obi, and no it isn't obvious whether the rtl-vs-expectations is sufficiently checked already.

I have made updates attempting to address this now.
Added new assertions to check the pma module's reported flags (buf, cache, err) vs expectations.

sva/cv32e40x_mpu_sva.sv
silabs-robin: I added a condition for IS_INSTR_SIDE, so that data-side should require these signals to be exactly equal.

But I am confused about how to have good excuse clauses.
There really is a problem with `norm || excuse` because the excuse might be too forgiving (e.g. `excuse==1`), and it is true that `norm ^ excuse` would prevent this problem, as far as I can understand.
But... when it is difficult to make an `excuse` that is disjoint from `norm`, then using `norm ^ (!norm && excuse)` (as I have done) might not necessarily be better. It might be, but I haven't yet been able to prove either way which case is true.

silabs-robin: Done.

TL;DR:
I have updated the obi-checking assertions and hope the assertions are satisfactory.

--------

I found this difficult to write in a satisfactory trustable manner.
Hypothetical/potential failures should not be masked.
`norm ^ (!norm && excuse)` I strongly suspect is meaningless; it should be boolean reduceable to `norm || excuse`.
Also, the existing cases are not actually `a->b, !a->!b` but rather `a->b, !a->c`, because merely each obi signal got inversed but not the entire expression (it actually CEXes if one tries to do `!a->!b` on e.g. cacheable).

I do think proper equivalence could guard against overly kind excuse clauses:
```
excuse := 1
expect -> (actual + excuse)    # would mask actual
!expect -> !(actual + excuse)    # would alert
```
Although, there seem to be another problem lurking here:
```
actual := 0
expect -> (actual + excuse)    # can ignore actual
!expect -> !(actual + excuse)    # can ignore actual
```
But one can select a block output (obi signal) as the sole antecedent, because it is an actual and not a prediction.
And because of the thing with the free related cover, it should remove the problem of bypassing the actual.
```
# tieing expected or excuse will either fail outright or be unable to bypass the other
# tieing actual will be caught as related cover (vacuous)
actual -> expected * !excuse
!actual -> !(expected * !excuse)
# note that both predictions are on the same side
# note that this implication uses and instead of or
```

silabs-robin:
@Silabs-ArjanB and @silabs-oivind I have tried to address all your feedback.
[Here](https://github.com/openhwgroup/cv32e40x/pull/117/files#diff-112be9149d4ba5a6be6ecebf49b09c6ad8a90d1a4037a838d8a133be5058b7cfR164) is the part for Øivind's feedback.
There is one remark I didn't make any changes for, and that is https://github.com/openhwgroup/cv32e40x/pull/117#discussion_r662198584.
I wrote comments to some of the feedback, but all the handled threads are marked as "resolved".

silabs-oivind (approved):

PR #123, opened 2021-07-02
Factoring halt_ex into EX stage local instr_valid.

silabs-oysteink:
SEC must be checked, currently there is a very long runtime


Silabs-ArjanB (approved):
SEC study to be performed later

PR #126, opened 2021-07-02
Minstret updates to make clean ci_check

silabs-oysteink:
Rewrote performance counters to have proper write enable and next_value. (For use in RVFI)

Added support for minstret. Notethat this uses wb_valid which factors in data_rvalid_i.
Added stall in EX if minstret is READ in EX and any instruction is valid in WB.
Updated RVFI to include dret pc in WB model and not updating EX pc in case of misaligned LSU.

ci_check is clean with one warning during debug_test (comes from verif, not RTL)

Synthesis shows around 450 gates increase compared to last hash.
This is probably due to using wb_valid in counting retired instructions.

### DIFF
--- a/bhv/cv32e40x_rvfi.sv
+++ b/bhv/cv32e40x_rvfi.sv
@@ -488,7 +488,9 @@ module cv32e40x_rvfi
 
       //// EX Stage ////
       if (instr_ex_valid_i && wb_ready_i) begin
-        pc_wdata [STAGE_EX] <= branch_taken_ex_i ? branch_target_ex_i : pc_wdata[STAGE_ID];
+        pc_wdata [STAGE_EX] <= branch_taken_ex_i    ? branch_target_ex_i :
+                               !lsu_misaligned_ex_i ? pc_wdata[STAGE_ID] :
+                               pc_wdata[STAGE_EX];
         debug    [STAGE_EX] <= debug    [STAGE_ID];
         rs1_addr [STAGE_EX] <= rs1_addr [STAGE_ID];
         rs2_addr [STAGE_EX] <= rs2_addr [STAGE_ID];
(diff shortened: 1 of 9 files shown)


bhv/cv32e40x_rvfi.sv
```diff
@@ -537,7 +539,9 @@ module cv32e40x_rvfi
       end
 
       // Set expected next PC, half-word aligned
-      rvfi_pc_wdata <= (exception_in_wb_i) ? exception_target_wb_i & ~32'b1 : pc_wdata[STAGE_EX] & ~32'b1;
+      rvfi_pc_wdata <= (exception_in_wb_i) ? exception_target_wb_i & ~32'b1 :
+                       (is_dret_wb       ) ? csr_dpc_q_i :
```
silabs-halfdan, line 543: Would like to use the insn_dret signal from the ex_wb pipe here at some point. But not important.

silabs-halfdan (approved):

bhv/cv32e40x_rvfi.sv
```diff
@@ -488,7 +488,9 @@ module cv32e40x_rvfi
 
       //// EX Stage ////
       if (instr_ex_valid_i && wb_ready_i) begin
-        pc_wdata [STAGE_EX] <= branch_taken_ex_i ? branch_target_ex_i : pc_wdata[STAGE_ID];
+        pc_wdata [STAGE_EX] <= branch_taken_ex_i    ? branch_target_ex_i :
+                               !lsu_misaligned_ex_i ? pc_wdata[STAGE_ID] :
```
Silabs-ArjanB, line 492: The misaligned handling looks like it is masking an RTL bug

Silabs-ArjanB (approved):
Merging, but will file two issues:

- The stated area increase should not be needed for fixing of minstret
- The RVFI fix related to misaligned load/store really points at an RTL bug and should be fixed there instead

PR #131, opened 2021-07-02
Implemented B extensions Zbb and Zbs

silabs-halfdan:
Implemented Zbb and Zbs instructions

Changes are SEC clean with the top level parameter  B_EXT == NONE.


Silabs-ArjanB (approved):
Code will only serve as initial reference. It does not contain any sharing.

PR #132, opened 2021-07-05
Multiplier SVAs valid/ready update

silabs-robin:
Formal regression was failing, so I took the opportunity to fix the multiplier SVAs that appear outdated after valid/ready handshaking was introduced.
Three things are changed: 1) "s_insistent_valid" replaces antecedent of input stability assertions, 2) "a_check_result_constant" was similarly updated, 3) a_check_acc_mul_value_zero was qualified with valid_i.

NB! This needs review before merging. (doesn't seem like uarch multiplier section reflects latest changes?)

Question: Should we deal with those 2 TODOs too?

### DIFF
--- a/sva/cv32e40x_mult_sva.sv
+++ b/sva/cv32e40x_mult_sva.sv
@@ -52,8 +52,12 @@ module cv32e40x_mult_sva
   assign mul_result = $signed(op_a_i) * $signed(op_b_i);
   a_mul_result : // check multiplication result for MUL
     assert property (@(posedge clk) disable iff (!rst_n)
-                     (valid_i && (operator_i == MUL_M32)) |->
-                     (result_o == mul_result)) else `uvm_error("mult", "MUL result check failed")
+                     (valid_i && (operator_i == MUL_M32)) |-> (result_o == mul_result))
+      else `uvm_error("mult", "MUL result check failed")
+  a_mul_valid : // check that MUL result is immediately qualified by valid_o
+    assert property (@(posedge clk) disable iff (!rst_n)
+                     (valid_i && (operator_i == MUL_M32)) |-> valid_o)
+      else `uvm_error("mult", "MUL result wasn't immediately valid")
 
 
   // Check result for all MULH flavors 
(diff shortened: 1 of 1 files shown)


sva/cv32e40x_mult_sva.sv
```diff
@@ -84,9 +85,10 @@ module cv32e40x_mult_sva
...
 
-  // Check that multiplier inputs are not changed in the middle of a MULH operation
-  logic         ready;
-  assign ready = ready_o && ready_i;
+  sequence s_insistent_valid;
+    @(posedge clk)
+    (valid_i && !ready_o) ##1 valid_i;
+  endsequence
 
   // TODO:low This assertion will soon be purposely broken as valid_i could be deasserted to abort ongoing multicyle instructions
```
Silabs-ArjanB, line 93: Can you fix the TODOs as well now that you are updating this file?

silabs-robin: I have fixed the 2 TODOs

----

> assign mulh_result_valid = valid_i && (operator_i == MUL_H) && ready_o; // TODO:low could valid_o be used directly here?

Not only "could" but I think it should/must.
`ready_o` before and after introducing ready/valid handshake protocol are identical in name, not identical in semantics.
Previously, it meant both "ready for new upstream inputs" _and_ "output can be used".
I think the only reason the assertions still worked is because incidentally `ready_o` is implied by some conditions relating to `valid_o`.
`valid_o` is the qualifier of outputs.

> This assertion [a_enable_constant_when_mulh_active] will soon be purposely broken

Previously `enable_i` had a stability requirement.
`valid_i` does not have such a requirement (afaik).
Hence I deleted this stale commented-away assertion.

----

I also added two new assertions.
If these are objectionable then I can remove them again.
First, `valid_o |-> valid_i` because mult doesn't register its inputs and the output validity therefore demands input validity.
Secondly, `(valid_o && ready_i) |-> ready_o` because the result has been consumed and we don't want to waste performance.

silabs-oivind (changes requested):
Please add check for valid_o in a_mul_result. It should be set immediately on single cycle multiplications.

silabs-robin:
> Please add check for valid_o in a_mul_result. It should be set immediately on single cycle multiplications.

I added that now.
Although I split it into a new assertion because it is supposedly easier for the tool.

silabs-oivind (approved):

PR #136, opened 2021-07-13
Updated RVFI branch probe

silabs-halfdan:
Updated RVFI branch probe to handle the jump-once update in the controller.

This fixes an issue where pc_wdata would not be set correctly (leading to rvfi_intr getting set) when a branch was taken during a bus stall.

### DIFF
--- a/bhv/cv32e40x_wrapper.sv
+++ b/bhv/cv32e40x_wrapper.sv
@@ -290,7 +290,7 @@ bind cv32e40x_sleep_unit:
          .lsu_type_id_i            ( core_i.id_stage_i.lsu_type                                           ),
          .lsu_we_id_i              ( core_i.id_stage_i.lsu_we                                             ),
 
-         .branch_taken_ex_i        ( core_i.controller_i.controller_fsm_i.branch_taken_ex                 ),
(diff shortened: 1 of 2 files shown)


bhv/cv32e40x_wrapper.sv
```diff
@@ -290,7 +290,7 @@ bind cv32e40x_sleep_unit:
          .lsu_type_id_i            ( core_i.id_stage_i.lsu_type                                           ),
          .lsu_we_id_i              ( core_i.id_stage_i.lsu_we                                             ),
 
-         .branch_taken_ex_i        ( core_i.controller_i.controller_fsm_i.branch_taken_ex                 ),
+         .branch_taken_ex_i        ( core_i.controller_i.controller_fsm_i.branch_in_ex                    ),
```
Silabs-ArjanB, line 293: Please rename branch_taken_ex_i to branch_in_ex_i in the RVFI module; now it is confusing.

Silabs-ArjanB (changes requested):

bhv/cv32e40x_wrapper.sv
silabs-halfdan: Good point, will update now.

Silabs-ArjanB (approved):

PR #158, opened 2021-07-30
Moved decision of legal CSR accesses from decoder to cs_registers.

silabs-oysteink:
Also killing CSR writes with csr_en instead of changing csr_op.
EX stage will deassert write enables and set ex_wb_pipe.illegal_insn in case of illegal CSR instructions.

SEC equivalent under the following assumption:
- No illegal CSR instructions (csr_en would stay high until EX, causing stalls)


rtl/cv32e40x_cs_registers.sv
```diff
@@ -179,12 +186,27 @@ module cv32e40x_cs_registers import cv32e40x_pkg::*;
   assign csr_waddr = csr_num_e'(ex_wb_pipe_i.csr_addr);
   assign csr_wdata = ex_wb_pipe_i.csr_wdata;
 
-  //TODO:OK We should have a better way for killing CSR insn other than forcing csr_op to CSR_OP_READ (csr_en already exists in pipeline)
-  assign csr_op       =  (!ctrl_fsm_i.kill_wb && ex_wb_pipe_i.instr_valid) ? ex_wb_pipe_i.csr_op : CSR_OP_READ;
+  assign csr_op    =  ex_wb_pipe_i.csr_op;
+
+  // CSR write operations in WB, actual csr_we_int may still become 1'b0 in case of CSR_OP_READ
+  assign csr_we    = (ex_wb_pipe_i.instr_valid && ex_wb_pipe_i.csr_en) && !ctrl_fsm_i.kill_wb;
```
Silabs-ArjanB, line 192: Why do you want this 'we' to ever be 1 for non-writes?

Silabs-ArjanB: Can you rename this signal to csr_en_gated and use a local valid signal as in the other pipeline stages?

Silabs-ArjanB (changes requested):
See comments

rtl/cv32e40x_core.sv
```diff
@@ -608,6 +609,8 @@ module cv32e40x_core import cv32e40x_pkg::*;
     // Interface to CSRs (SRAM like)
     .csr_rdata_o                ( csr_rdata              ),
 
+    .csr_illegal_o              (csr_illegal_ex          ),
```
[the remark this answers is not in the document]
silabs-oysteink: Done

rtl/cv32e40x_cs_registers.sv
silabs-oysteink: Changed to csr_en_gated

Silabs-ArjanB (approved):

PR #169, opened 2021-08-09
Implemented the Zbc B extension

silabs-halfdan:
Implemented  the Zbc B extension.

Sequentially equivalent for B_EXT=NONE and B_EXT=ZBA_ZBB_ZBS. Enabled with B_EXT=ZBA_ZBB_ZBC_ZBS.

### DIFF
--- a/rtl/include/cv32e40x_pkg.sv
+++ b/rtl/include/cv32e40x_pkg.sv
@@ -119,15 +119,21 @@ typedef enum logic [ALU_OP_WIDTH-1:0]
  ALU_B_SH3ADD = 6'b100110, // funct3 = 110
 
  // B, Zbb
- ALU_B_CLZ    = 6'b100000, // (funct3 = 001)
- ALU_B_CTZ    = 6'b101000, // (funct3 = 001)
- ALU_B_CPOP   = 6'b100011, // (funct3 = 001)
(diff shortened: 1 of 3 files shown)


rtl/include/cv32e40x_pkg.sv
```diff
@@ -119,15 +119,21 @@ typedef enum logic [ALU_OP_WIDTH-1:0]
  ALU_B_SH3ADD = 6'b100110, // funct3 = 110
 
  // B, Zbb
- ALU_B_CLZ    = 6'b100000, // (funct3 = 001)
- ALU_B_CTZ    = 6'b101000, // (funct3 = 001)
- ALU_B_CPOP   = 6'b100011, // (funct3 = 001)
+ ALU_B_CLZ    = 6'b100000, // funct3 = 001
```
Silabs-ArjanB, line 122: Can you put the brackets back around the funct3 comments? Itried to make the 3 lower bits of this type equal to funct3 and I used brackets to denote the cases where this was no longer possible.

Silabs-ArjanB (changes requested):

rtl/include/cv32e40x_pkg.sv
silabs-halfdan: Ah, that makes sense. Will fix it now.

Silabs-ArjanB (approved):

PR #173, opened 2021-08-10
Made B_EXT a top level parameter

silabs-halfdan:

### DIFF
--- a/bhv/cv32e40x_wrapper.sv
+++ b/bhv/cv32e40x_wrapper.sv
@@ -42,6 +42,7 @@ module cv32e40x_wrapper
   import cv32e40x_pkg::*;
 #(
   parameter NUM_MHPMCOUNTERS             =  1,
+  parameter B_EXT                        = NONE,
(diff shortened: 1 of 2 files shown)


bhv/cv32e40x_wrapper.sv
```diff
@@ -42,6 +42,7 @@ module cv32e40x_wrapper
   import cv32e40x_pkg::*;
 #(
   parameter NUM_MHPMCOUNTERS             =  1,
+  parameter B_EXT                        = NONE,
```
Silabs-ArjanB, line 45: This should have the b_ext_e type as well

Silabs-ArjanB (changes requested):

bhv/cv32e40x_wrapper.sv
silabs-halfdan: ah yes...

Silabs-ArjanB (approved):

PR #177, opened 2021-08-11
RVFI debug updates

silabs-halfdan:
Added debug mode and debug cause signal, renamed rvfi_dbg signal to rvfi_debug to match spec.

### DIFF
--- a/bhv/cv32e40x_rvfi.sv
+++ b/bhv/cv32e40x_rvfi.sv
@@ -65,6 +65,7 @@ module cv32e40x_rvfi
    input logic [31:0]                         pc_wb_i,
    input logic                                wb_ready_i,
    input logic                                wb_valid_i,
+   input logic                                ebreak_in_wb_i,
    input logic [31:0]                         instr_rdata_wb_i,
    input logic                                exception_in_wb_i,
    // Register writes
@@ -486,7 +487,8 @@ module cv32e40x_rvfi
         // IF stage is killed and not valid during debug entry. If debug is taken,
         // debug cause is saved to propagate through rvfi pipeline together with next valid instruction
         if (debug_taken_if) begin
-          debug_cause_if_next <= debug_cause_i; // debug cause input only valid during debug taken
+          // Debug cause input only valid during debug taken
+          debug_cause_if_next <=  ebreak_in_wb_i ? 3'h1 : debug_cause_i; // Debug entry from debug mode caused by EBREAK is not captured by debug_cause_i
         end
       end
 
--- a/bhv/cv32e40x_rvfi.sv
+++ b/bhv/cv32e40x_rvfi.sv
@@ -488,7 +488,9 @@ module cv32e40x_rvfi
         // debug cause is saved to propagate through rvfi pipeline together with next valid instruction
         if (debug_taken_if) begin
           // Debug cause input only valid during debug taken
-          debug_cause_if_next <=  ebreak_in_wb_i ? 3'h1 : debug_cause_i; // Debug entry from debug mode caused by EBREAK is not captured by debug_cause_i
+          // Special case for debug entry from debug mode caused by EBREAK as it is not captured by debug_cause_i
+          // A higher priority debug request (e.g. trigger match) will pull ebreak_in_wb_i low and allow the debug cause to propagate
+          debug_cause_if_next <=  ebreak_in_wb_i ? 3'h1 : debug_cause_i;
         end
       end
 
--- a/bhv/cv32e40x_rvfi.sv
+++ b/bhv/cv32e40x_rvfi.sv
@@ -80,6 +80,11 @@ module cv32e40x_rvfi
    input logic [31:0]                         exception_target_wb_i,
    input logic [31:0]                         mepc_target_wb_i,
 
+   input                                      PrivLvl_t current_priv_lvl_i,
+
+   input logic                                debug_mode_i,
+   input logic [2:0]                          debug_cause_i,
+
    //// CSR Probes ////
    input                                      Status_t csr_mstatus_n_i,
    input                                      Status_t csr_mstatus_q_i,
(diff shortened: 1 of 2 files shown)


bhv/cv32e40x_rvfi.sv
```diff
@@ -80,6 +80,11 @@ module cv32e40x_rvfi
    input logic [31:0]                         exception_target_wb_i,
    input logic [31:0]                         mepc_target_wb_i,
 
+   input                                      PrivLvl_t current_priv_lvl_i,
+
+   input logic                                debug_mode_i,
+   input logic [2:0]                          debug_cause_i,
```
Silabs-ArjanB, line 86: debug_cause_i is not available in the RTL so it should not be an input here; it should be constructed in the RVFI module

bhv/cv32e40x_rvfi.sv
```diff
@@ -157,16 +162,21 @@ module cv32e40x_rvfi
    output logic [ 0:0]                        rvfi_intr,
    output logic [ 1:0]                        rvfi_mode,
    output logic [ 1:0]                        rvfi_ixl,
-   output logic [ 0:0]                        rvfi_dbg, // TODO: Rename to rvfi_debug
 
+   output logic [ 0:0]                        rvfi_debug,
```
Silabs-ArjanB, line 166: Duplicate signal; now called rvfi_dbg_mode

bhv/cv32e40x_rvfi.sv
```diff
@@ -157,16 +162,21 @@ module cv32e40x_rvfi
...
+   output logic [ 0:0]                        rvfi_dbg_mode,
```
Silabs-ArjanB, line 167: Remove [0:0]

bhv/cv32e40x_rvfi.sv
```diff
@@ -439,8 +448,9 @@ module cv32e40x_rvfi
       ex_mem_addr        <= '0;
       ex_mem_wdata       <= '0;
       ex_csr_rdata       <= '0;
-
-      rvfi_dbg           <= '0;
+      rvfi_debug         <= '0;
```
Silabs-ArjanB, line 451: Remove rvfi_debug everywhere

bhv/cv32e40x_rvfi.sv
```diff
@@ -538,6 +547,12 @@ module cv32e40x_rvfi
         rvfi_mem_wmask <= mem_wmask[STAGE_EX];
         rvfi_mem_addr  <= ex_mem_addr;
         rvfi_mem_wdata <= ex_mem_wdata;
+
+        rvfi_mode      <= current_priv_lvl_i;
+
+        rvfi_debug     <= debug[STAGE_EX];
+        rvfi_dbg_mode  <= debug_mode_i;
+        rvfi_dbg_cause <= debug_cause_i;
```
Silabs-ArjanB, line 555: Cause needs to be constructed in the RVFI module

bhv/cv32e40x_rvfi.sv
```diff
@@ -538,6 +547,12 @@ module cv32e40x_rvfi
         rvfi_mem_wmask <= mem_wmask[STAGE_EX];
         rvfi_mem_addr  <= ex_mem_addr;
         rvfi_mem_wdata <= ex_mem_wdata;
+
+        rvfi_mode      <= current_priv_lvl_i;
+
+        rvfi_debug     <= debug[STAGE_EX];
+        rvfi_dbg_mode  <= debug_mode_i;
```
Silabs-ArjanB, line 554: Discuss with Øystein whether 'debug mode' needs to get constructed inside RVFI or whether you can import it from the RTL.

Silabs-ArjanB (changes requested):
See comments

bhv/cv32e40x_rvfi.sv
silabs-halfdan: ctrl_fsm.debug_cause contains the debug cause. Discussed the use of this signal with Øystein and using it should be ok as long as it is only sampled during the DEBUG_TAKEN state of the controller FSM.

bhv/cv32e40x_rvfi.sv
silabs-halfdan: Discussed this with Øystein, probing it in IF to allow any un-killed LSU instructions to complete will give us the correct result.

bhv/cv32e40x_rvfi.sv
silabs-halfdan: Will update these signals.

bhv/cv32e40x_rvfi.sv
silabs-halfdan: Will do

bhv/cv32e40x_rvfi.sv
Silabs-ArjanB: If you do this, please add clear explanation in your code.

bhv/cv32e40x_rvfi.sv
Silabs-ArjanB: I am not sure about whether these signals are the same; sent you a more detailed note about it.

bhv/cv32e40x_rvfi.sv
```diff
@@ -464,42 +477,54 @@ module cv32e40x_rvfi
...
+      if (if_valid_i && id_ready_i) begin
+        debug_cause[STAGE_IF]<= debug_cause_if_next;
+        debug_cause_if_next  <= '0;
+        debug_mode[STAGE_IF] <= debug_mode_i; // Probing in IF to make sure any LSU instructions that are not killed can complete
+      end else begin
+        // IF stage is killed and not valid during debug entry. If debug is taken,
+        // debug cause is saved to propagate through rvfi pipeline together with next valid instruction
+        if (debug_taken_if) begin
+          // Debug cause input only valid during debug taken
+          debug_cause_if_next <=  ebreak_in_wb_i ? 3'h1 : debug_cause_i; // Debug entry from debug mode caused by EBREAK is not captured by debug_cause_i
```
silabs-oysteink, line 491: Won't this give the wrong cause is debug was entered due to an external request (code 0x3) while at the same time an ebreak (with lower priority) was in WB? I would think we might need to qualify with checking if we are in debug mode as well.

silabs-halfdan: An ebreak in wb should be killed (suppressing the ebreak_in_wb signal) if we get an external request at the same time.

silabs-oysteink: > An ebreak in wb should be killed (suppressing the ebreak_in_wb signal) if we get an external request at the same time.

I think this was addresses at my first (then edited) comment? Ebreak are in fact a higher priority debug cause than an external haltreq (I mistakenly thought the other way around in my first comment), so in that case the ebreak would 'win'.  I think the only concern was the trigger match, which as mentioned will suppress any special instruction bits.

silabs-halfdan: These comments got a bit messy with edits, we discussed it offline and will leave this code as is.

silabs-oysteink: Maybe just add some more comments to clarify why it is safe, ie ebreak with trigger match will never be signalled as ebreak_in_wb for instance.

silabs-halfdan: Ok, will add a comment

silabs-oysteink (approved):

bhv/cv32e40x_rvfi.sv
```diff
@@ -538,6 +564,11 @@ module cv32e40x_rvfi
         rvfi_mem_wmask <= mem_wmask[STAGE_EX];
         rvfi_mem_addr  <= ex_mem_addr;
         rvfi_mem_wdata <= ex_mem_wdata;
+
+        rvfi_mode      <= current_priv_lvl_i;
```
Silabs-ArjanB, line 568: Can you add a todo here to indicate that an explanation (or fix) is needed to explain why current_priv_lvl_i is valid when sampled here? (Now this signal is constant, so it doesn't matter, but this might likely break when user mode is added)

Silabs-ArjanB (changes requested):

bhv/cv32e40x_rvfi.sv
silabs-halfdan: Good point, I'll add it now.

Silabs-ArjanB (approved):

PR #179, opened 2021-08-12
Bugfixes for kill_wb during debug mode and flops using wrong clock

silabs-oysteink:
Bugfix: Some flip flops in the controller were using the ungated clock by mistake.

Bugfix: kill_wb during DEBUG_TAKEN would kill regular ebreaks (withouth dcsr.ebreakm)
if debug was caused by an external request. Ebreaks were also killed during debug mode.
This fix does not affect RTL, but affects RVFI/RVVI.


Silabs-ArjanB (approved):

silabs-oysteink:
This PR is SEC equivalent when excluding the changes to the flipflops using the wrong clock.

Issue #187, opened 2021-08-17
RVFI may set dpc before entering debug handler during single-step mode


strichmo:
### Component
**Component:RVFI**
**Component:Doc **: For issues in the Documentation (e.g. for User Manual or README.md files)
**Component:Other**: For any other issues

### Steps to Reproduce
Please provide:
1. e422fa37d0f5c50e526dbfddb01f927c7d1fd513 (cv32e40x/dev branch)
2. Using Xcelium 21.07.a001
```
%  makeuvmt test TEST=debug_test WAVES=1 ADV_DEBUG=1 SEED=1680260209 CV_CORE=cv32e40x
```

The RVFI will report the dpc CSR updating to the "next" instruction during the instruciton retirement of the stepped instruction (0xca94).  However the ISS will always wait until the first instruction retirement in the debug handler before updating dpc to the next instruction (0xca98).  This behavior is intermittent-not all single-stepped instructions show this behavior on RVFI.

![image](https://user-images.githubusercontent.com/56401023/129760575-d853d940-0033-4c9e-94f2-5b1e977b2b4a.png)

silabs-halfdan:
@silabs-oysteink will implement a fix for this in RTL.

Silabs-ArjanB:
Hi @silabs-oysteink , this also touches on the discussion we just had. Here you could also argue that all instructions should set dpc as part of their execution in single step mode.

silabs-oysteink:
Agreed @Silabs-ArjanB, there will be a PR shortly which fixes this and also aligns on CSR update when the stepped instruction retires.

Silabs-ArjanB:
Should be fixed with https://github.com/openhwgroup/cv32e40x/pull/192

strichmo:
Confirmed fixed.  The debug_test is passing robustly with and without random stalls.

PR #192, opened 2021-08-19
Single step fix (wb_valid)

silabs-oysteink:
Single step would be signalled as pending whenever an instruction reached WB stage.

This could cause LSU instructions with late rvalid to finish after the controller
entered debug mode. This fix waits until the actual wb_valid to signal a pending
single step.

Fixes issue #187

### DIFF
--- a/rtl/cv32e40x_controller_fsm.sv
+++ b/rtl/cv32e40x_controller_fsm.sv
@@ -146,6 +146,7 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
 
   logic [4:0] exc_cause; // id of taken interrupt
 
+
   // Mux selector for vectored IRQ PC
   assign ctrl_fsm_o.m_exc_vec_pc_mux = (mtvec_mode_i == 2'b0) ? 5'h0 : exc_cause;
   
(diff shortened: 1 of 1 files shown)


rtl/cv32e40x_controller_fsm.sv
```diff
@@ -214,10 +215,11 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
   // Single step will need to finish insn in WB, including LSU
   // Need to check for finished multicycle instructions, avoiding rvalid (would cause path to instr_o)
   // todo: better way of factoring in last part of a misaligned LSU instruction
+  // todo:ok: May factor away single_step_allowed completely once PMA/LSU support is merged
   assign single_step_allowed = !(id_ex_pipe_i.lsu_misaligned && id_ex_pipe_i.instr_valid) && !obi_data_req_q;
                              
   // Single step are mutually exclusive from any other reason to enter debug
-  assign pending_single_step = (!debug_mode_q && debug_single_step_i && ex_wb_pipe_i.instr_valid) && !pending_debug;
+  assign pending_single_step = (!debug_mode_q && debug_single_step_i && wb_valid_i) && !pending_debug;
```
Silabs-ArjanB, line 222: This assumes that wb_valid_i is only 1 at the end of a multi-operation instruction. This is how it currently is for misaligned loads/stores; we should make sure this stays that way for push/pop/etc.

Silabs-ArjanB (approved):

PR #193, opened 2021-08-20
LSU PMA support

silabs-oysteink:
PMA/MPU work by silabs-oivind, controller and plumbing by silabs-oysteink.

- Implements PMA check for misaligned load/store to IO regions
- Implements support for LSU PMA exceptions.
- Suppresses ex_wb_pipe.rf_we for first half of misaligned LSU instructions (only second half will write to RF)
- Suppresses WB stage rf_we in case of PMA errors.

### DIFF
--- a/rtl/cv32e40x_wb_stage.sv
+++ b/rtl/cv32e40x_wb_stage.sv
@@ -84,7 +84,8 @@ module cv32e40x_wb_stage import cv32e40x_pkg::*;
   // Note that write back is not suppressed during bus errors (in order to prevent
   // a timing path from the late arriving data_err_i into the register file).
   //
-  // Note that the register file is written twice in case of a misaligned load.
+  // Note that the register file is only written for the last part of a misaligned load.
+  // (rf_we suppressed in ex_stage for the first part)
   //
   // Note that the register file is written multiple times in case waited loads (in
   // order to prevent a timing path from the late arriving data_rvalid_i into the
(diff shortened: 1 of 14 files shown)


silabs-oivind (approved):

rtl/cv32e40x_wb_stage.sv
```diff
@@ -84,13 +85,17 @@ module cv32e40x_wb_stage import cv32e40x_pkg::*;
...
   //
   // Note that the register file is written multiple times in case waited loads (in
   // order to prevent a timing path from the late arriving data_rvalid_i into the
   // register file.
+  //
+  // In case of MPU/PMA error, the register file should not be written.
+  // rf_we_wb_o is deasserted if lsu_mpu_status is not equal to MPU_OK
 
-  assign rf_we_wb_o     = ex_wb_pipe_i.rf_we && instr_valid ; // TODO:OK:low deassert in case of MPU error (already do this in EX stage)
+  assign rf_we_wb_o     = ex_wb_pipe_i.rf_we && (lsu_mpu_status_i == MPU_OK) && instr_valid;
```
Silabs-ArjanB, line 98: How is it guaranteed that mpu status is okay for non-load-store instructions? Can an assertion be added for that or (if this is not true) can this signal be qualified?

Silabs-ArjanB (changes requested):

Silabs-ArjanB (approved):

PR #194, opened 2021-08-20
Cvxif user interface

GuillaumeChauvon:
Hello, 

@Silabs-ArjanB @davideschiavone @JeanRochCoulon @vogelpi @treforsouthwell

I added a user interface to the specification. 

This interface allows to communicate user data from the core to the coprocessor and to obtain a possible response. 
It gives a new possibility to communicate with the coprocessor for custom purposes. 

It will make the build of an interconnect for identical coprocessor much easier. 
I also believe it can have much more use for example : 
- Set or modify parameters in the coprocessor without instruction
- Access a coprocessor register without instruction

Regards, 
Guillaume Chauvon
Intern at Thales DIS Design Services


Silabs-ArjanB:
Hi @GuillaumeChauvon, thank you for this pull request. The proposal only adds interface signals and does not add any semantics for them (other than the validity with respect to the handshake signals). As such (and as I communicated earlier) I think this should be left outside the spec. If Thales wants to add some custom signals and custom CSRs, then of course they can do that, but mandating signals without any semantics is only going to lead to more questions and unclarity about what processors and/or coprocessors are required to support.

GuillaumeChauvon:
Hi @Silabs-ArjanB, thank you for your reply. You are right. 
After thinking it through we might be able to implement our use case (multiple custom coprocessor) without this user interface. If it is mandatory to our use case then we will add it on our own side.

Silabs-ArjanB:
Will not merge this

PR #199, opened 2021-08-25
Clarify handling of misaligned load/stores in CVXIF

vogelpi:
It has been agreed to:
- Split misaligned load/stores inside the core always.
- To delay the memory request/response handshake until its clear whether a synchronous exception is triggered or not by that request.
- Keep the memory request signals stable until the handshake as been
  performed (no retraction of `x_mem_valid_i` without handshake).


vogelpi:
FYI @Silabs-ArjanB

docs/user_manual/source/x_ext.rst
```diff
@@ -562,6 +568,7 @@ The only rule related to valid and ready signals is that:
 Specifically note the following:
 
 * The valid signal is allowed to be retracted (e.g. in case that the related instruction is killed in the core's pipeline before the corresponding ready is signaled).
+  This does not hold for ``x_mem_valid_i``: once asserted it must remain asserted until the handshake with ``x_mem_ready_o`` has been performed.
```
Silabs-ArjanB, line 571: My original intention (although not very clearly stated I see now) was that the retraction of valid signals is only allowed by the CPU, never by the coprocessor. So, if I had worded that more clearly this additional remark should not be needed. Why specifically did you add this line?

vogelpi: I wasn't aware that only the CPU was allowed to retract the valid. But this makes sense.

The reason for adding that sentence the potentially delayed handshake for misaligned load/store requests: the first aligned request might be sent out to the bus already in the first cycle (before the CPU sends the ready). So it might be too late for the co-processor to de-assert the valid as the actual load/store is already ongoing.

Silabs-ArjanB: Okay, understood. I will merge this and then later clarify that a coprocessor is never allowed to retract handshakes. Thanks for the help!

Silabs-ArjanB (approved):

PR #213, opened 2021-09-03
Added RVFI support for 40s features

silabs-halfdan:
Added usermode and pmp csrs for cv32e40s support. Tied off these features for the 40x.


bhv/cv32e40x_rvfi.sv
```diff
@@ -568,15 +599,16 @@ module cv32e40x_rvfi
         rvfi_mem_addr  <= ex_mem_addr;
         rvfi_mem_wdata <= ex_mem_wdata;
 
-        rvfi_mode      <= current_priv_lvl_i; // todo: Verify / explain this when user mode is implemented
+        // Separate privelege level signal needed for LSU intructions, both valid in WB
+        rvfi_mode      <= lsu_en_wb_i ? priv_lvl_lsu_i :  priv_lvl_i;
```
Silabs-ArjanB, line 603: Why is there a need for a separate LSU privilege level signal? Does this have to do with the mprv bit and if so, where is it documented in the RVFI spec that that bit needs to be taken into account?

silabs-halfdan: Yes, it does. I can add a comment here and a note in the reference manual.

Silabs-ArjanB: I am not sure if this is the logical thing to do as opposed to just exporting the mode in which the instruction is executed (without taking mprv into account). Did you check with Claire what is the intention of this signal?

Silabs-ArjanB (approved):
Will approve this and at the same time file a ticket to judge if the interpretation of the rvfi_mode signal is correct or not.

PR #214, opened 2021-09-06
Removed WAW stall.

silabs-oysteink:
We now have a single write port to the register file, and this hazard is obsolete.

NOT SEC clean.


silabs-oysteink:
Passes ci_check

Silabs-ArjanB (approved):

PR #215, opened 2021-09-07
Fencei handshake

silabs-oivind:
ci_check OK, but tb needs to be updated to drive fencei_flush_ack_i.

### DIFF
--- a/rtl/cv32e40x_controller_fsm.sv
+++ b/rtl/cv32e40x_controller_fsm.sv
@@ -86,7 +86,11 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
   input  logic        id_ready_i,       // ID stage is ready for new data
   input  logic        ex_valid_i,       // EX stage has valid (non-bubble) data for next stage
   input  logic        wb_ready_i,       // WB stage is ready for new data,
-  input  logic        wb_valid_i        // WB stage ha valid (non-bubble) data
+  input  logic        wb_valid_i,       // WB stage ha valid (non-bubble) data
+
+  // Fencei flush handshake
+  output logic        fencei_flush_req_o,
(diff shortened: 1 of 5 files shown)


sva/cv32e40x_controller_fsm_sva.sv
```diff
@@ -143,6 +156,21 @@ module cv32e40x_controller_fsm_sva
                     (ctrl_fsm_o.kill_wb) |-> (!csr_we_i) )
     else `uvm_error("controller", "csr written while kill_wb is asserted")
 
+  a_kill_wb_ex_id_if:
```
Silabs-ArjanB, line 159: Use disable iff (!rst_n) for all assertions (please also update the ones in this file that had them missing as well)

rtl/cv32e40x_controller_fsm.sv
```diff
@@ -467,14 +478,17 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
...
+            // Kill if,id and ex stages, halt wb stage
             ctrl_fsm_o.kill_if = 1'b1;
             ctrl_fsm_o.kill_id = 1'b1;
             ctrl_fsm_o.kill_ex = 1'b1;
-            ctrl_fsm_o.pc_set  = 1'b1;
-            ctrl_fsm_o.pc_mux  = PC_FENCEI;
+            ctrl_fsm_o.halt_wb = 1'b1;
 
-            //TODO:OK:low Drive fence.i interface
+            if(fencei_ready) begin
```
Silabs-ArjanB, line 488: Did you take into account 'single stepping a fencei' instruction?

Silabs-ArjanB (changes requested):

PR #216, opened 2021-09-07
Updated register file to a different data type to the register file wrapper

silabs-halfdan:
Updated register file to not use same data type as register file wrapper to allow inserting ECC in other cores (like 40s) while keeping the register file merge-able.

### DIFF
--- a/rtl/cv32e40x_register_file.sv
+++ b/rtl/cv32e40x_register_file.sv
@@ -27,24 +27,24 @@
 module cv32e40x_register_file import cv32e40x_pkg::*;
 (
     // Clock and Reset
-    input  logic         clk,
-    input  logic         rst_n,
+    input logic                           clk,
+    input logic                           rst_n,
 
     // Read ports
-    input  rf_addr_t     raddr_i [REGFILE_NUM_READ_PORTS],
-    output rf_data_t     rdata_o [REGFILE_NUM_READ_PORTS],
+    input                                 rf_addr_t raddr_i [REGFILE_NUM_READ_PORTS],
+    output logic [REGFILE_WORD_WIDTH-1:0] rdata_o [REGFILE_NUM_READ_PORTS],
 
     // Write ports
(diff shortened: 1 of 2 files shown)


rtl/cv32e40x_register_file.sv
```diff
@@ -27,24 +27,24 @@
...
     // Clock and Reset
-    input  logic         clk,
-    input  logic         rst_n,
+    input logic                           clk,
+    input logic                           rst_n,
 
     // Read ports
-    input  rf_addr_t     raddr_i [REGFILE_NUM_READ_PORTS],
-    output rf_data_t     rdata_o [REGFILE_NUM_READ_PORTS],
+    input                                 rf_addr_t raddr_i [REGFILE_NUM_READ_PORTS],
```
Silabs-ArjanB, line 34: Please align type same as on other lines

rtl/cv32e40x_register_file.sv
```diff
@@ -27,24 +27,24 @@
...
+    output logic [REGFILE_WORD_WIDTH-1:0] rdata_o [REGFILE_NUM_READ_PORTS],
 
     // Write ports
-    input rf_addr_t      waddr_i [REGFILE_NUM_WRITE_PORTS],
-    input rf_data_t      wdata_i [REGFILE_NUM_WRITE_PORTS],
-    input logic          we_i [REGFILE_NUM_WRITE_PORTS]
+    input                                 rf_addr_t waddr_i [REGFILE_NUM_WRITE_PORTS],
```
Silabs-ArjanB, line 38: Please align type same as on other lines

Silabs-ArjanB (changes requested):
Good idea to do this change already in the 40x; just fix the alignment

rtl/cv32e40x_register_file.sv
silabs-halfdan: This is the emacs standard for typedef-ed signals, but I agree its prettier to align with the basic data types. Will push an update.

Silabs-ArjanB (approved):

PR #221, opened 2021-09-10
Added support for NMI.

silabs-oysteink:
SEC clean if data_err_i is assumed to be 1'b0;

### DIFF
--- a/rtl/cv32e40x_controller_fsm.sv
+++ b/rtl/cv32e40x_controller_fsm.sv
@@ -57,8 +57,8 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
   input  logic        obi_data_req_i,             // LSU OBI interface req
 
   // From WB stage
-  input  logic        lsu_err_wb_i,               // LSU caused bus_error in WB stage
-  input  logic [31:0] lsu_addr_wb_i,              // LSU address in WB stage
+  input  logic        lsu_err_wb_i,               // LSU caused bus_error in WB stage, gated with data_rvalid_i inside load_store_unit
+  input  logic [31:0] lsu_addr_wb_i,              // LSU address in WB stage //todo: only needed if MTVAL is implemented
 
   // From LSU (WB)
   input  mpu_status_e lsu_mpu_status_wb_i,        // MPU status (WB timing)
@@ -98,6 +98,10 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
   // Sticky version of debug_req_i
   logic debug_req_q;
 
+  // Sticky version of lsu_err_wb_i
+  logic nmi_pending_q;
+  logic nmi_is_store_q; // 1 for store, 0 for load
+
   // Debug mode
   logic debug_mode_n;
   logic debug_mode_q;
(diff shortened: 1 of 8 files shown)


sva/cv32e40x_core_sva.sv
```diff
@@ -109,30 +109,35 @@ always_ff @(posedge clk , negedge rst_ni)
       // code needs to check priority of what to expect
       if (!first_illegal_found && ex_wb_pipe.instr_valid && !irq_ack_o && !(ctrl_pending_debug && ctrl_debug_allowed) &&
         !(ex_wb_pipe.instr.bus_resp.err || (ex_wb_pipe.instr.mpu_status != MPU_OK)) &&
+        !(ctrl_fsm.exc_pc_mux == EXC_PC_NMI) &&
```
Silabs-ArjanB, line 112: Why do these lines need to be added (whereas regular interrupts are not excluded in a similar manner)?

rtl/cv32e40x_controller_fsm.sv
```diff
@@ -252,16 +258,19 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
...
+  assign pending_interrupt = irq_req_ctrl_i && !debug_mode_q;
 
   // Allow interrupts to be taken only if there is no data request in WB, 
   // and no data_req has been clocked from EX to environment.
   // LSU instructions which were suppressed due to previous exceptions or trigger match
   // will be interruptable as they were convered to NOP in ID stage.
   // TODO:OK:low May allow interuption of Zce to idempotent memories
   assign interrupt_allowed = !(ex_wb_pipe_i.lsu_en && ex_wb_pipe_i.instr_valid) && !obi_data_req_q && !debug_mode_q;
-                               
+  assign nmi_allowed = interrupt_allowed;
```
Silabs-ArjanB, line 272: The user manual states the following:

NMIs are never masked by the MIE bit. NMIs are masked however while in debug mode or while single stepping with STEPIE = 0 in the dcsr CSR.

Where is this STEPIE behavior implemented? We should also add this for regular IRQs (although at the moment, due to the relative priority between debug and IRQ that is probably a don't care).

Silabs-ArjanB (changes requested):

Silabs-ArjanB:
Can you also document the use of nmi_addr_i in the interrupt related section?

sva/cv32e40x_core_sva.sv
silabs-oysteink: The regular interrupts are excluded by checking for !irq_ack_o two lines above. If not added, illegal instructions (and the other types in the same block of code) could be detected while a higher priority NMI was pending and taken.

Silabs-ArjanB: Okay, I missed that. So this means then that irq_ack_o is not set for NMIs, correct? (That would be yet another reason to remove irq_ack_o)

rtl/cv32e40x_controller_fsm.sv
silabs-oysteink: The MIE bit is handled in the int_controller, and thus the fsm_controller will never see interrupts pending from int_controller if the MIE bit is cleared. This factors into the 'pending_interrupt' signal, and not the interrupt_allowed.

For single stepping, I was under the assumption that NMIs should never be disabled during step, disregarding dcsr.stepie. I will go back and check the debug spec for this.

sva/cv32e40x_core_sva.sv
silabs-oysteink: That's right, it's only set for regular interrupts.

Silabs-ArjanB (approved):

PR #238, opened 2021-09-21
Corrected fencei description

Silabs-ArjanB:


silabs-robin (approved):
I think those changes made it clearer. Looks correct too.

PR #241, opened 2021-09-23
Cleanup of code triggering warnings in formal verif

silabs-hfegran:
Signed-off-by: Henrik Fegran <Henrik.Fegran@silabs.com>

### DIFF
--- a/bhv/cv32e40x_core_log.sv
+++ b/bhv/cv32e40x_core_log.sv
@@ -47,6 +47,7 @@ module cv32e40x_core_log import cv32e40x_pkg::*;
   
 );
 
+`ifndef FORMAL
   // Log top level parameter values
   initial
   begin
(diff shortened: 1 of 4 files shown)


Silabs-ArjanB (commented):
I understand you don't like the 'initial'; can you however confirm that the corrected code can actually get triggered?

silabs-hfegran:
> I understand you don't like the 'initial'; can you however confirm that the corrected code can actually get triggered?

"Initial"-blocks never gets triggered in formal - formal tool ignores the code block altogether (with a warning).
and the updated variant is verified in sim and formal with correct and incorrect atomic/cacheable settings.
The check a_pma_valid_config is difficult to test, as a mismatch of PMA_NUM_REGIONS and actual defined regions will result in a compile-time failure, passes correctly as long as the regions are correctly set. 

The other initial block is just log prints which will never be used in, and also ignored by formal.

PR #243, opened 2021-09-24
RVFI: Fixed lint warnings and fixed issue with pma register assignment

silabs-halfdan:

### DIFF
--- a/bhv/cv32e40x_rvfi.sv
+++ b/bhv/cv32e40x_rvfi.sv
@@ -870,9 +870,9 @@ module cv32e40x_rvfi
   generate
     for (genvar i = 0; i < 16; i++ ) begin // Max 16 pmp regions
       // 4 regions in each register
-      assign rvfi_csr_wdata_d.pmpcfg[i/4][i%4+:4] = csr_pmpcfg_n_i[i];
-      assign rvfi_csr_rdata_d.pmpcfg[i/4][i%4+:4] = csr_pmpcfg_q_i[i];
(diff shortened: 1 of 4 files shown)


bhv/cv32e40x_rvfi.sv
```diff
@@ -884,6 +884,11 @@ module cv32e40x_rvfi
   assign rvfi_csr_rdata_d.pmpmseccfg[0]   = csr_pmpmseccfg_q_i;
   assign rvfi_csr_wmask_d.pmpmseccfg[0]   = csr_pmpmseccfg_we_i ? '1 : '0;
 
+  // pmpmseccfg[1] not supported
```
Silabs-ArjanB, line 887: - Can you add a todo here as this CSR should actually actually exist, but might be always 0; either way it should be connected to the RTL and not tied off here.
- Also add a todo to change pmpmseccfg[0] into mseccfg and pmpmseccfg[1] into mseccfgh (which are the official names for these CSRs) in both RVFI and the RTL (or just do the renaming now)

Silabs-ArjanB (changes requested):

bhv/cv32e40x_rvfi.sv
silabs-halfdan: Will make some small updates and add todos.

silabs-halfdan: Implementing these small changes instead of adding todos here. Will add todo in 40s for mseccfgh probe hookup.

Silabs-ArjanB (approved):

PR #247, opened 2021-09-29
Bugfix: Did not stall for reading of instret, only minstret

silabs-oysteink:
Stalling EX on all performance counter reads if there is a valid instruction in WB.

### DIFF
--- a/rtl/cv32e40x_controller_bypass.sv
+++ b/rtl/cv32e40x_controller_bypass.sv
@@ -54,8 +54,7 @@ module cv32e40x_controller_bypass import cv32e40x_pkg::*;
   input  logic        debug_trigger_match_id_i,         // Trigger match in ID
 
   // From EX
-  input  csr_num_e    csr_raddr_ex_i,             // CSR read address (EX)
-  input  logic        csr_counter_read_i,         // CSR is reading a counter (EX)
+  input  logic        csr_counter_read_i,         // CSR is reading a counter (EX).
 
   // From WB
   input  logic        wb_ready_i,                 // WB stage is ready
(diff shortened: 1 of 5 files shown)


rtl/cv32e40x_controller_bypass.sv
```diff
@@ -55,6 +55,7 @@ module cv32e40x_controller_bypass import cv32e40x_pkg::*;
 
   // From EX
   input  csr_num_e    csr_raddr_ex_i,             // CSR read address (EX)
+  input  logic        csr_counter_read_i,         // CSR is reading a counter (EX)
```
Silabs-ArjanB, line 58: Remove csr_raddr_ex_i pin if no longer used

rtl/cv32e40x_controller_bypass.sv
```diff
@@ -194,8 +187,8 @@ module cv32e40x_controller_bypass import cv32e40x_pkg::*;
       ctrl_byp_o.csr_stall = 1'b1;
     end
 
-    // Stall (EX) due to minstret read
-    if (minstret_read_in_ex && ex_wb_pipe_i.instr_valid) begin
+    // Stall (EX) due to performance counter read
+    if (csr_counter_read_i && ex_wb_pipe_i.instr_valid) begin
```
Silabs-ArjanB, line 191: It does not look correct that the qualification with (id_ex_pipe_i.instr_valid && id_ex_pipe_i.csr_en) has been removed. I suggest that you reintroduce the minstret_read_in_ex signal. Also add a note that the stalling might be conservative (as we would also stall for counter reads not involving minstret)

Silabs-ArjanB (changes requested):

rtl/cv32e40x_controller_bypass.sv
silabs-oysteink: Removed

rtl/cv32e40x_controller_bypass.sv
silabs-oysteink: csr_counter_read_i is derived from csr_raddr inside cs_registers. The read address is gated to 0 if no valid csr instruction is in EX.

Silabs-ArjanB (approved):

PR #248, opened 2021-09-29
New event for minstret counter.

silabs-oysteink:
Instead of using wb_valid which factors in data_rvalid_i,
minstret is counted for the first cycle of a valid instruction in WB.
For LSU instructions with stalls on rvalid, this means we count before the instruction retires.
RVFI is altered to detect this and report correct values for the retired instruction.

### DIFF
--- a/rtl/cv32e40x_controller_fsm.sv
+++ b/rtl/cv32e40x_controller_fsm.sv
@@ -55,6 +55,7 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
   input  id_ex_pipe_t id_ex_pipe_i,        
   input  logic        branch_decision_ex_i,       // branch decision signal from EX ALU
   input  logic        obi_data_req_i,             // LSU OBI interface req
+  input  logic        lsu_split_ex_i,             // LSU is splitting misaligned, first half is in EX
 
   // From WB stage
   input  logic        lsu_err_wb_i,               // LSU caused bus_error in WB stage, gated with data_rvalid_i inside load_store_unit
@@ -159,6 +160,8 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
   logic       fencei_flush_req_set;
   logic       fencei_req_and_ack_q;
   logic       fencei_ongoing;    
+
+  logic       wb_minstret_event;
 
   assign fencei_ready = 1'b1; // TODO: connect when write buffer is implemented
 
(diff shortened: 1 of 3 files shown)


rtl/cv32e40x_controller_fsm.sv
```diff
@@ -801,6 +804,25 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
...
+      end else begin
+        // Clear event to make sure only first cycle of WB counts
+        // unless WB is halted (for fence.i for example), then we must wait until un-halt to clear event flag
+        if(!ctrl_fsm_o.halt_wb) begin
+          wb_minstret_event <= 1'b0;
+        end
+      end
+    end
+  end
+
```
silabs-oivind, line 825: This will be used for multiple events, not only minstret. Could you pick a more generic name? E.g. wb_valid_event ?
Could also create a common signal to be used in the counter events, factoring in kill_wb and halt_wb.
Should wb_minstret_event also be cleared if WB is killed?

silabs-oysteink: Good point, I'll rename it and make a common signal factoring in kill and halt. I think the flag should be cleared if the instruction in WB is killed, as it should not increase any counters post mortem.

silabs-halfdan (approved):

silabs-halfdan (approved):

rtl/cv32e40x_controller_fsm.sv
```diff
@@ -801,6 +816,25 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
...
+      wb_counter_event <= 1'b0;
+    end else begin
+      // Minstret event for first cycle of WB
+      // First half of misaligned split will not count
+      if(ex_valid_i && wb_ready_i && !lsu_split_ex_i) begin
+        wb_counter_event <= 1'b1;
+      end else begin
+        // Clear event to make sure only first cycle of WB counts
+        // unless WB is halted (for fence.i for example), then we must wait until un-halt to clear event flag
+        if(!ctrl_fsm_o.halt_wb) begin
```
Silabs-ArjanB, line 831: This code is not intuitive to me. Would replacing this with if (!wb_ready_i) give the same result (as I think that code would be more intuitive). Specifically this is not clear because it does not explain how double counting is prevented for stalled loads/stores and it also makes fence.i a special case which I don't think is needed.

Silabs-ArjanB (changes requested):

Silabs-ArjanB (approved):

silabs-oysteink:
Self merging after updating comments and aligning with Arjan.

Issue #231, opened 2021-09-16
Make irq_ack/id internal and add dbg_ack


Silabs-ArjanB:
- Make irq_ack_o, irq_id_o local in the top level
- Remove all mentions of above signals from the User Manual (no need to add the core local signals in the User Manual)
- Add 'debug ack' pulse (and implement similar to IRQ ack in controller FSM)
- core-v-verif or the wrapper will have to probe these signals into the top level of the core (discuss with Steve what is best)

Further requirements (to be checked with assertions (some have to be at wrapper level I expect):
 
- NMI will not be reported on irq_ack
- Every irq_ack pulse must lead to rvfi_intr (but rvfi_intr is allowed to become 1 without a preceding irq_ack (in case of synchronous exceptions)
- Every dbg_ack pulse must lead to rvfi_dbg <> 0x0
- Every rvfi_dbg <> 0x0 must be preceded with dbg_ack pulse

cv32e40x_core.sv will look like the following:

module cv32e40x_core (
…
)
 
  logic                   irq_ack;                               // Used (only) by verification environment
  logic [4:0]          irq_id;                                   // Used (only) by verification environment
  logic                   dbg_ack;                             // Used (only) by verification environment
 
  assign irq_ack   = ctrl_fsm.irq_ack;          // Used (only) by verification environment (do not remove)
  assign irq_id      = ctrl_fsm.irq_id;                // Used (only) by verification environment (do not remove)
  assign dbg_ack = ctrl_fsm.dbg_ack;      // Used (only) by verification environment (do not remove)
 
  …
 
endmodule

PR #251, opened 2021-10-01
Fix github issue #231

silabs-oivind:
Make irq_ack/id internal and add dbg_ack


Silabs-ArjanB (commented):
Can you also completely remove these irq signals from the User Manual, for example from https://cv32e40x-user-manual.readthedocs.io/en/latest/integration.html ?

silabs-oivind:
> Can you also completely remove these irq signals from the User Manual, for example from https://cv32e40x-user-manual.readthedocs.io/en/latest/integration.html ?

Done

Silabs-ArjanB (approved):

PR #255, opened 2021-10-04
Timing opt: JR hazard detection

silabs-oysteink:
Added separate hazard detection for JR. JR will always read from RF, so the hazard logic does not include checks for rf_re when checking ID vs EX and WB for RF address.

SEC equivalent.

### DIFF
--- a/rtl/cv32e40x_controller_bypass.sv
+++ b/rtl/cv32e40x_controller_bypass.sv
@@ -64,7 +64,9 @@ module cv32e40x_controller_bypass import cv32e40x_pkg::*;
 );
 
   logic [REGFILE_NUM_READ_PORTS-1:0] rf_rd_ex_match;
+  logic [REGFILE_NUM_READ_PORTS-1:0] rf_rd_ex_jr_match;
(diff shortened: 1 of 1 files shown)


rtl/cv32e40x_controller_bypass.sv
```diff
@@ -64,7 +64,9 @@ module cv32e40x_controller_bypass import cv32e40x_pkg::*;
 );
 
   logic [REGFILE_NUM_READ_PORTS-1:0] rf_rd_ex_match;
+  logic [REGFILE_NUM_READ_PORTS-1:0] rf_rd_ex_jr_match;
```
Silabs-ArjanB, line 67: The jr match should actually only be defined for RS1, so it should not be a vector. Could you make it 1 bit wide and pull it outside the generate loop (now we have one bit in the vector for which we likely get a lint warning)

Silabs-ArjanB (changes requested):

rtl/cv32e40x_controller_bypass.sv
silabs-oysteink: Updated

Silabs-ArjanB (approved):

PR #260, opened 2021-10-05
Timing fix: Performance counter event not flopped

silabs-oysteink:
Performance counter event for jr_stall was not flopped, leading to bad timing.


silabs-oysteink:
Self servicing this simple bugfix

PR #265, opened 2021-10-08
Redirect extension interface into submodules

michael-platzer:
This PR splits the modports of the extension interface into individual modports for the `compressed`, `issue`, `commit`, `mem`, `mem_result`, and `result` interfaces, which allows to wire these ports to the respective units within CV32E40X.

### DIFF
--- a/rtl/cv32e40x_id_stage.sv
+++ b/rtl/cv32e40x_id_stage.sv
@@ -19,6 +19,7 @@
 //                 Davide Schiavone - pschiavo@iis.ee.ethz.ch                 //
 //                 Halfdan Bechmann - halfdan.bechmann@silabs.com             //
 //                 Øystein Knauserud - oystein.knauserud@silabs.com           //
+//                 Michael Platzer - michael.platzer@tuwien.ac.at             //
 //                                                                            //
 // Design Name:    Instruction Decode Stage                                   //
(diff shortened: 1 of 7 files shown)


rtl/cv32e40x_id_stage.sv
```diff
@@ -89,7 +90,11 @@ module cv32e40x_id_stage import cv32e40x_pkg::*;
   // Stage ready/valid
   output logic        id_ready_o,     // ID stage is ready for new data
   output logic        id_valid_o,     // ID stage has valid (non-bubble) data for next stage
-  input  logic        ex_ready_i      // EX stage is ready for new data
+  input  logic        ex_ready_i,     // EX stage is ready for new data
+
+  // eXtension interface
+  if_core_v_xif.cpu_issue  if_xif_issue,
+  if_core_v_xif.cpu_commit if_xif_commit
```
Silabs-ArjanB, line 97: It is our expectation that the commit interface will need to be connected to the 'controller FSM' (https://github.com/openhwgroup/cv32e40x/blob/master/rtl/cv32e40x_controller_fsm.sv) as that is the module (with WB timing) that decides which instructions are allowed to be committed and which ones get killed.

Silabs-ArjanB (changes requested):
See comment about commit interface

PR #273, opened 2021-10-13
Trigger uses PC from IF stage

silabs-oysteink:
Changed trigger match PC comparison to use IF stage PC instead of ID stage.

SEC clean


rtl/cv32e40x_core.sv
```diff
@@ -350,6 +350,8 @@ module cv32e40x_core import cv32e40x_pkg::*;
 
     .dpc_i               ( dpc                       ), // debug return address
 
+    .trigger_match_i     ( debug_trigger_match_if    ),
```
Silabs-ArjanB, line 353: Pin and signal names do not match; either use trigger_match everywhere or debug_trigger_match (remark holds for pins, signals and also the field in the pipe registers)

PR #282, opened 2021-10-15
Implement X_EXT, instruction IDs, and stalls in case of data hazards for the CORE-V-XIF

michael-platzer:
This PR continues implementation of the CORE-V-XIF eXtension interface by adding the `X_EXT` parameter to enable/disable the XIF, instruction IDs, and stalls in case there is a data hazard for an XIF instruction in the EX stage.

### DIFF
--- a/rtl/cv32e40x_id_stage.sv
+++ b/rtl/cv32e40x_id_stage.sv
@@ -34,6 +34,7 @@ module cv32e40x_id_stage import cv32e40x_pkg::*;
 #(
   parameter A_EXTENSION             =  0,
   parameter b_ext_e B_EXT           =  NONE,
+  parameter bit     X_EXT           =  0,
   parameter DEBUG_TRIGGER_EN        =  1
 )
 (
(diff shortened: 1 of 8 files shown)


rtl/cv32e40x_id_stage.sv
```diff
@@ -398,7 +401,7 @@ module cv32e40x_id_stage import cv32e40x_pkg::*;
 
     // Register file control signals
     .rf_re_o                         ( rf_re                     ),
-    .rf_we_o                         ( rf_we                     ),
+    .rf_we_o                         ( rf_we_dec                 ),
```
Silabs-ArjanB, line 404: Just FYI: In principle we do not want such signal renaming (except for the _o removal and typically the addition of _id, _ex, etc.). We will see how to best clean that up later (to maximize code reuse with the CV32E40S) as I understand you somehow need to merge in the XIF we signal, so you need to rename a signal somewhere. No need to change anything at this point in time.

rtl/cv32e40x_id_stage.sv
```diff
@@ -616,50 +624,75 @@ module cv32e40x_id_stage import cv32e40x_pkg::*;
...
-    if (rst_n == 1'b0) begin
-      xif_accepted_q <= 1'b0;
-      xif_rejected_q <= 1'b0;
-    end else begin
-      xif_accepted_q <= !(id_valid_o && ex_ready_i) && xif_insn_accept;
-      xif_rejected_q <= !(id_valid_o && ex_ready_i) && xif_insn_reject;
-    end
-  end
+  generate
+    if (X_EXT) begin
```
Silabs-ArjanB, line 628: We have a convention that generate blocks have a name (that eases debug and probing into scopes as it prevents automatically generated names). Can you change this into:

if (X_EXT) begin: x_ext

rtl/cv32e40x_id_stage.sv
```diff
@@ -616,50 +624,75 @@ module cv32e40x_id_stage import cv32e40x_pkg::*;
...
-  assign xif_issue_if.x_issue_req.frs_valid = '0;
+      // an instruction was offloaded successfully if the coprocessor accepts it (or has accepted it)
+      assign xif_insn_accept = (xif_issue_if.x_issue_valid && xif_issue_if.x_issue_ready &&  xif_issue_if.x_issue_resp.accept) || xif_accepted_q;
+      assign xif_insn_reject = (xif_issue_if.x_issue_valid && xif_issue_if.x_issue_ready && !xif_issue_if.x_issue_resp.accept) || xif_rejected_q;
+
+      assign xif_we = xif_issue_if.x_issue_valid && xif_issue_if.x_issue_resp.writeback;
 
-  // need to wait if the coprocessor is not ready and has not already accepted or rejected the instruction
-  assign xif_waiting = xif_issue_if.x_issue_valid && !xif_issue_if.x_issue_ready && !xif_accepted_q && !xif_rejected_q;
+    end else begin
```
Silabs-ArjanB, line 678: and this into:

end else begin: no_x_ext

Can you also please do the same for the other generate statements you introduced?

rtl/cv32e40x_if_stage.sv
```diff
@@ -298,8 +305,28 @@ instruction_obi_i
...
+  //---------------------------------------------------------------------------
+
+  generate
+    if (X_EXT) begin
+
+      // TODO: implement offloading of compressed instruction
+      assign xif_compressed_if.x_compressed_valid = '0;
+      assign xif_compressed_if.x_compressed_req   = '0;
+
+      assign xif_id = id_offload_i ? if_id_pipe_o.xif_id + 1 : if_id_pipe_o.xif_id;
```
Silabs-ArjanB, line 320: The IDs just need to be unique, they do not necessarily need to be consecutive, so the usage of id_offload_i might not be needed. We do somehow need to enforce the uniqueness of course, but that will tie into how many offloads we allow to be outstanding at a given moment. Can you just add a todo that we assert that the oustanding IDs are unique? Also, if you think we can do without id_offload_i, please remove it. If is stays then it needs to be renamed so that we do not have the little bit unpredictable renaming that currently happens in :

.id_offload_i ( xif_issue_if.x_issue_valid

So maybe:

.issue_valid_i ( xif_issue_if.x_issue_valid

Silabs-ArjanB (changes requested):

rtl/cv32e40x_wb_stage.sv
```diff
@@ -156,6 +156,7 @@ module cv32e40x_wb_stage import cv32e40x_pkg::*;
 
   // TODO: How to handle conflicting values of ex_wb_pipe_i.rf_waddr and xif_result_if.x_result.rd?
   // TODO: How to handle conflicting values of ex_wb_pipe_i.rf_we (based on xif_issue_if.x_issue_resp.writeback in ID) and xif_result_if.x_result.we?
+  // TODO: Check whether result IDs match the instruction IDs propagated along the pipeline
 
   // Need to wait for the result
   assign xif_waiting = ex_wb_pipe_i.instr_valid && ex_wb_pipe_i.xif_en && !xif_result_if.x_result_valid;
```
Silabs-ArjanB, line 162: @strichmo We asked Michael to implement the top level X_EXT parameter to indicate whether the X interface is present or not. If X_EXT is 0, then the interface will not be functional, but the pins will of course still be present. In that scenario, do you care about code like this? Do you want such code to be inside generate statements (for code coverage reasons) as much as possible? Also, we can mandate that if X_EXT = 0 that the integrator connects the related top level pins to 0. Should we avoid relying on such correct tie-off or do you prefer that we use the parameter to make sure that nothing bad happens in case of an incorrect tie off (I would prefer that).

rtl/cv32e40x_id_stage.sv
michael-platzer: Sure, will add this.

rtl/cv32e40x_if_stage.sv
michael-platzer: If the IDs do not need to be consecutive, then I can of course remove this. Maybe I misinterpreted the spec, but I thought that following sentence implies consecutive IDs:

> The `id` values of subsequent commit transactions will increment (and wrap around)

Silabs-ArjanB: You are right about that. I wrote that myself, but I missed it when looking back in the spec today. There is no real need for this however, but let us for now keep your implementation then. The only justification for the requirement would be to allow for easier assertions to perform some checks on the interfaces.

Silabs-ArjanB (approved):
Thanks Michael!

PR #284, opened 2021-10-15
Implemented write buffer (1/2)

halfdan-dolva:
Implemented write buffer on transfer request signals.
Does not currently lead to any performance increase as we still wait for rvalid regardless of bufferable.
The change is logically equivalent when assumed 1 is set on data_gnt.

### DIFF
--- a/rtl/cv32e40x_write_buffer.sv
+++ b/rtl/cv32e40x_write_buffer.sv
@@ -104,7 +104,9 @@ module cv32e40x_write_buffer
   ///////////////////////////////////////////
   // Outputs
   ///////////////////////////////////////////
-  assign ready_o = (state == WBUF_FULL) ? (bufferable && ready_i) : (bufferable || ready_i);
+  assign ready_o = (state == WBUF_FULL) ?
+                   (bufferable && ready_i) : // A downstream ready will free up the buffer for a new bufferable transfer
+                   (bufferable || ready_i);  // Ready for bufferable transfer, accept any transfer if downstream is ready
   assign valid_o = (state == WBUF_FULL) || valid_i;
   assign trans_o = (state == WBUF_FULL) ? trans_q : trans_i;
 
--- a/sva/cv32e40x_write_buffer_sva.sv
+++ b/sva/cv32e40x_write_buffer_sva.sv
@@ -103,19 +103,17 @@ module cv32e40x_write_buffer_sva
   property p_data_q_value;
     @(posedge clk) disable iff (!rst_n)
       bufferable && (state === WBUF_EMPTY) && (valid_i && !ready_i) ||
-      bufferable && (state === WBUF_FULL)  && (valid_i &&  ready_i) |=> (trans_q === $past(trans_i));
+      bufferable && (state === WBUF_FULL)  && (valid_i &&  ready_i) |=> (trans_q === $past(trans_i)) && (state === WBUF_FULL);
   endproperty : p_data_q_value
 
   a_data_q_value: assert property (p_data_q_value)
     else `uvm_error("write_buffer", "trans_q is should have been same as trans_i from previous edge");
 
-  // When trans_q is changed, the state should transit from EMPTY to FULL or from FULL to FULL
+  // When trans_q is changed,
   property p_data_q_condition;
     @(posedge clk) disable iff (!rst_n)
-      // ##1 used to avoid first clock edge when past data_q value is unknown
-      ##1 (trans_q !== $past(trans_q)) |-> $past(bufferable) && (state === WBUF_FULL) && $past(valid_i) &&
-                                          (($past(state) === WBUF_EMPTY) && !$past(ready_i) ||
-                                           ($past(state) === WBUF_FULL)  &&  $past(ready_i) );
+      !(bufferable && (state === WBUF_EMPTY) && (valid_i && !ready_i) ||
+        bufferable && (state === WBUF_FULL)  && (valid_i &&  ready_i)) |=> $stable(trans_q);
       endproperty : p_data_q_condition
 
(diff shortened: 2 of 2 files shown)


rtl/cv32e40x_write_buffer.sv
```diff
@@ -0,0 +1,107 @@
...
+      trans_q <= '0;
+    end else if (push) begin
+      trans_q <= trans_i;
+    end
+  end
+
+  ///////////////////////////////////////////
+  // Outputs
+  ///////////////////////////////////////////
+  assign ready_o = (state == WBUF_EMPTY) && (bufferable || ready_i);
```
Silabs-ArjanB, line 102: This does not look optimal to me. What would happen for example if there are two bufferable stores in a row and at the end of a first cycle the first store actually enters the write buffer (i.e. state becomes full) and in the second cycle the initial store receives its grant? I would expect that at the end of this second cycle the second bufferable store can enter the write buffer (but it would not).

sva/cv32e40x_write_buffer_sva.sv
```diff
@@ -0,0 +1,138 @@
...
+      waited_addr_ph <= valid_o && !ready_i;
+    end
+  end
+
+  ///////////////////////////////////////////
+  // Verify OBI protocol adherence
+  ///////////////////////////////////////////
+  // verify that valid_o is held during waited addr phase
+  property p_valid_o_held_during_waited_addr_ph;
+    @(posedge clk)
```
Silabs-ArjanB, line 61: Please always use disable iff (!rst_n) for all assertions

sva/cv32e40x_write_buffer_sva.sv
```diff
@@ -0,0 +1,138 @@
...
+  logic               waited_addr_ph;
+  always_ff @(posedge clk, negedge rst_n) begin
+    if(!rst_n) begin
+      waited_addr_ph <= 1'b0;
+    end else begin
+      waited_addr_ph <= valid_o && !ready_i;
+    end
+  end
+
+  ///////////////////////////////////////////
```
Silabs-ArjanB, line 56: Do you expect that both interfaces of the write buffer are OBI compliant? If so, add assertions on both interfaces.

rtl/cv32e40x_write_buffer.sv
```diff
@@ -0,0 +1,107 @@
...
+    push       = 1'b0;
+
+    case(state)
+      WBUF_EMPTY: begin // buffer is empty
+        if(bufferable && valid_i && !ready_i) begin
+          next_state = WBUF_FULL;
+          push       = 1'b1; // push incoming data into the buffer
+        end
+      end
+      WBUF_FULL: begin // buffer is full
```
Silabs-ArjanB, line 80: I expect that you will need to allow for a FULL-> FULL transition

rtl/cv32e40x_write_buffer.sv
Silabs-ArjanB: I will re-look at the assertions after the design has been updated.

silabs-hfegran (changes requested):

sva/cv32e40x_write_buffer_sva.sv
```diff
@@ -0,0 +1,138 @@
...
+      bufferable && (state === WBUF_EMPTY) && (valid_i && !ready_i) |=> (trans_q === $past(trans_i));
+  endproperty : p_data_q_value
+
+  a_data_q_value: assert property (p_data_q_value)
+    else `uvm_error("write_buffer", "trans_q is should have been same as trans_i from previous edge");
+
+  // When trans_q is changed, the state should transit from EMPTY to FULL
+  property p_data_q_condition;
+    @(posedge clk)
+      ##1 (trans_q !== $past(trans_q)) |-> (state === WBUF_FULL) && ($past(bufferable) && ($past(state) === WBUF_EMPTY) && ($past(valid_i) && !$past(ready_i))); // ##1 used to avoid first clock edge when past data_q value is unknown
```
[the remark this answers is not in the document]
silabs-hfegran: (same goes for the updated property, with minor modifications)

sva/cv32e40x_write_buffer_sva.sv
```diff
@@ -99,18 +125,18 @@ module cv32e40x_write_buffer_sva
...
 
   a_ready_o_one: assert property (p_ready_o_one)
-    else `uvm_error("write_buffer", "When ready_o is 1, state should be EMPTY");
+    else `uvm_error("write_buffer", "When ready_o is 1, state should be EMPTY or OBI should be ready");
 
   // ready_o should be 0 only when state is FULL if the transfer is bufferable
   property p_ready_o_zero;
-    @(posedge clk)
+    @(posedge clk) disable iff (!rst_n)
       bufferable && !ready_o |-> (state === WBUF_FULL);
```
Silabs-ArjanB, line 140: The right hand side can be made stronger

rtl/cv32e40x_write_buffer.sv
```diff
@@ -99,7 +104,7 @@ module cv32e40x_write_buffer
   ///////////////////////////////////////////
   // Outputs
   ///////////////////////////////////////////
-  assign ready_o = (state == WBUF_EMPTY) && (bufferable || ready_i);
+  assign ready_o = (state == WBUF_FULL) ? (bufferable && ready_i) : (bufferable || ready_i);
```
Silabs-ArjanB, line 107: This deserves some comments

silabs-halfdan: Will add some comments

rtl/cv32e40x_write_buffer.sv
silabs-halfdan: This arc is now added

PR #306, opened 2021-11-08
Updated assertions for debug single step + interrupt taken.

silabs-oysteink:

### DIFF
--- a/sva/cv32e40x_controller_fsm_sva.sv
+++ b/sva/cv32e40x_controller_fsm_sva.sv
@@ -168,6 +168,7 @@ module cv32e40x_controller_fsm_sva
     else `uvm_error("controller", "LSU instruction follows WFI")
 
   // Check that no instructions are valid in ID or EX when a single step is taken
+  // In case of interrupt during step, ID/EX validity must be checked by checking for kill_id/ex
   // Exception if first phase of a misaligned LSU gets an MPU error, then 
   // the controller will kill the pipeline and jump to debug with dpc set to exception handler,
   // while id_ex_pipe may still contain the valid last phase of the misaligned LSU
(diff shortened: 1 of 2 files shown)


sva/cv32e40x_controller_fsm_sva.sv
```diff
@@ -176,7 +177,8 @@ module cv32e40x_controller_fsm_sva
     assert property (@(posedge clk) disable iff (!rst_n)
             (pending_single_step && (ctrl_fsm_ns == DEBUG_TAKEN) &&
             (lsu_mpu_status_wb_i == MPU_OK))
-            |-> (!id_ex_pipe_i.instr_valid && !if_id_pipe_i.instr_valid))
+            |-> ((!id_ex_pipe_i.instr_valid && !if_id_pipe_i.instr_valid) ||
```
Silabs-ArjanB, line 180: Did the original assertion fail?

silabs-robin: @Silabs-ArjanB the original does indeed fail in formal. But, @silabs-oysteink , do you know what changed that made it not pass anymore?

silabs-oysteink: The assertion shall check that the pipeline is empty when a single step becomes pending.
When allowing interrupts to set "pending_single_step", the stepped instruction could be in _any_ pipeline stage, and would get killed by the taken interrupt. The check is done in the same cycle as the kill, and thus the pipeline would not be empty as seen from the registered instr_valid signals. Adding this check to ensure we don't enter debug with stray/un-killed instructions left in the pipeline.

sva/cv32e40x_core_sva.sv
```diff
@@ -280,13 +280,13 @@ always_ff @(posedge clk , negedge rst_ni)
...
+  // Single step with interrupt goes straight to debug witout retiring
+  // any instructions before entering debug mode
   a_single_step_with_irq :
     assert property (@(posedge clk) disable iff (!rst_ni)
-                      (inst_taken && dcsr.step && !ctrl_fsm.debug_mode && interrupt_taken) [*1:2]
-                      ##1 inst_taken [->1]
-                      |-> (ctrl_fsm.debug_mode && dcsr.step))
+                      (dcsr.step && !ctrl_fsm.debug_mode && interrupt_taken)
+                      ##0 !wb_valid ##1 !wb_valid
+                      |=> (ctrl_fsm.debug_mode && dcsr.step))
```
silabs-robin, line 289: @silabs-oysteink I don't know all timing details here so could you explain why there is a check for `!wb_valid`, why there is a check for `##1 !wb_valid` and why there is `|=>` instead of `|->`?

silabs-oysteink: The interrupt taken will be asserted during the state FUNCTIONAL.  The first ##0 checks that we don't retire any instruction while we take the interrupt. The second (##1) check that we don't retire any while in DEBUG_TAKEN state. And finally the debug_mode flop will go to 1 one cycle after DEBUG_TAKEN.

Silabs-ArjanB: Please add such clarification as comments in the code.

Silabs-ArjanB (approved):

silabs-robin (approved):

silabs-oysteink:
Self-merging as both reviewers approved.

PR #307, opened 2021-11-09
Fix MPU SVAs so the simulator can compile

silabs-robin:
Simulation failed to compile because it doesn't have support for "static cast" used while assigning signals in a bind.
Formal had no problem problem with it but simulation did.

With this PR, simulation can now compile, and the PMA assertions still pass formal for all 7 pma configs.

Note, the write buffer related PMA assertions are planned to be updated once the write buffer implementation is finished, so these assignments are only temporary.


silabs-robin:
Covered in https://github.com/openhwgroup/cv32e40x/pull/308 instead

PR #310, opened 2021-11-09
add parameters to write_buffer RTL module for SVA

strichmo:
This avoids hierarchical parameters which are not supported in some simulators.  There should be no functional change introduced with this PR.

Fixes issues with Metrics/Dsim and latest RTL hash.  See this: https://github.com/openhwgroup/core-v-verif/pull/1010#issuecomment-964357559


silabs-oysteink (approved):

Silabs-ArjanB (approved):
Could have added a comment in the write buffer why the parameters are added (as they are not used for the RTL itself)

PR #314, opened 2021-11-12
Implemented response filter in lsu

silabs-halfdan:
Implemented response filter in the load/store unit for performance improvements to bufferable writes.

### DIFF
--- a/rtl/cv32e40x_load_store_unit.sv
+++ b/rtl/cv32e40x_load_store_unit.sv
@@ -52,7 +52,7 @@ module cv32e40x_load_store_unit import cv32e40x_pkg::*;
 
   // Stage 1 outputs (WB)
   output logic [31:0] lsu_addr_1_o,
-  output logic        lsu_err_1_o,           
+  output logic        lsu_err_1_o,
   output logic [31:0] lsu_rdata_1_o,            // LSU read data
   output mpu_status_e lsu_mpu_status_1_o,       // MPU (PMA) status, response/WB timing. To controller and wb_stage
 
@@ -84,7 +84,7 @@ module cv32e40x_load_store_unit import cv32e40x_pkg::*;
   logic [31:0]  resp_rdata;
   logic         resp_err;               // Unused for now
   data_resp_t   resp;
-  
+
   // Transaction request (from cv32e40x_mpu to cv32e40x_write_buffer)
   logic          buffer_trans_valid;
   logic          buffer_trans_ready;
@@ -103,7 +103,7 @@ module cv32e40x_load_store_unit import cv32e40x_pkg::*;
   // Transaction response (from cv32e40x_data_obi_interface to cv32e40x_mpu)
   logic           bus_resp_valid;
   obi_data_resp_t bus_resp;
-  
+
   // Counter to count maximum number of outstanding transactions
   logic [1:0]   cnt_q;                  // Transaction counter
   logic [1:0]   next_cnt;               // Next value for cnt_q
@@ -465,14 +465,14 @@ module cv32e40x_load_store_unit import cv32e40x_pkg::*;
   assign valid_1_o = (cnt_q == 2'b00) ? 1'b0 : last_q && resp_valid && valid_1_i; // todo:AB (cnt_q == 2'b00) should be same as !WB.lsu_en
 
   // LSU EX stage readyness requires two criteria to be met:
-  // 
+  //
   // - A data request has been forwarded/accepted (trans_valid && trans_ready)
   // - The LSU WB stage is available such that EX and WB can be updated in lock step
   //
   // Default (if there is not even a data request) LSU EX is signaled to be ready, else
   // if there are no outstanding transactions the EX stage is ready again once the transaction
   // request is accepted (at which time this load/store will move to the WB stage), else
-  // in case there is already at least one outstanding transaction (so WB is full) the EX 
+  // in case there is already at least one outstanding transaction (so WB is full) the EX
   // and WB stage can only signal readiness in lock step (so resp_valid is used as well).
 
   // todo:AB lsu_en_gated should maybe be replaced by valid_0_i
@@ -493,7 +493,7 @@ module cv32e40x_load_store_unit import cv32e40x_pkg::*;
   // otherwise we may let a new instruction into EX, overwriting second phase of split access..
   assign ready_0_o = done_0 && !lsu_split_0_o;
 
-  
+
   // Export mpu status to WB stage/controller
   assign lsu_mpu_status_1_o = resp.mpu_status;
 
@@ -502,11 +502,11 @@ module cv32e40x_load_store_unit import cv32e40x_pkg::*;
 
 
   //////////////////////////////////////////////////////////////////////////////
-  // Counter (cnt_q, next_cnt) to count number of outstanding OBI transactions 
+  // Counter (cnt_q, next_cnt) to count number of outstanding OBI transactions
   // (maximum = DEPTH)
-  // 
+  //
   // Counter overflow is prevented by limiting the number of outstanding transactions
-  // to DEPTH. Counter underflow is prevented by the assumption that resp_valid = 1 
+  // to DEPTH. Counter underflow is prevented by the assumption that resp_valid = 1
    // will only occur in response to accepted transfer request (as per the OBI protocol).
   //////////////////////////////////////////////////////////////////////////////
 
@@ -585,7 +585,7 @@ module cv32e40x_load_store_unit import cv32e40x_pkg::*;
   assign trans.prot[0]   = 1'b1;  // Transfers from LSU are data transfers
   assign trans.prot[2:1] = PRIV_LVL_M; // Machine mode
   assign trans.memtype   = 2'b00; // memtype is assigned in the MPU, tie off.
-  
+
   cv32e40x_mpu
   #(
     .IF_STAGE           ( 0                    ),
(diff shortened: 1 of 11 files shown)


rtl/cv32e40x_load_store_unit.sv
```diff
@@ -603,17 +610,37 @@ module cv32e40x_load_store_unit import cv32e40x_pkg::*;
     .core_resp_valid_o    ( resp_valid         ),
     .core_resp_o          ( resp               ),
 
-    .bus_trans_valid_o    ( buffer_trans_valid ),
-    .bus_trans_ready_i    ( buffer_trans_ready ),
-    .bus_trans_o          ( buffer_trans       ),
-    .bus_resp_valid_i     ( bus_resp_valid     ),
+     .bus_trans_valid_o    ( filter_trans_valid ),
```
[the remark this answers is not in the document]
Silabs-ArjanB: Make sure prefixes matches signal in a better way. In particular now it is confusing that bus_resp_i comes from a different place than the other bus_* signals.

rtl/cv32e40x_load_store_unit.sv
```diff
@@ -563,13 +569,14 @@ module cv32e40x_load_store_unit import cv32e40x_pkg::*;
     end else begin
       // Update for valid addresses if not blocked by controller
       if(!ctrl_fsm_i.block_data_addr && (trans_valid && trans_ready)) begin
-        lsu_addr_1_o <= trans.addr;
+        lsu_addr_1_o <= bus_trans.addr;
```
Silabs-ArjanB, line 572: Why are bus_trans* signals used together with trans* signals? Check maybe with Øystein as he had some todos here anyway.

rtl/cv32e40x_load_store_unit.sv
```diff
@@ -563,13 +569,14 @@ module cv32e40x_load_store_unit import cv32e40x_pkg::*;
...
       end
     end
   end
 
-  // Validate bus_error on rvalid (WB stage)
-  assign lsu_err_1_o = resp_valid && resp_err; // todo:low this gating is a bit weird; all LSU WB stage outputs should only be used when resp_valid = 1
+  // Validate bus_error on rvalid from the bus (WB stage)
+  // For bufferable transfers, this can happen many cycles after the pipeline control logic has seen the filtered resp_valid
+  assign lsu_err_1_o = bus_resp_valid && resp_err;
```
Silabs-ArjanB, line 579: Here signals from different places in the design are combined. It fixes the bug Øivind found, but it should be explained why it is okay to combine these two signals with different prefixes.

rtl/cv32e40x_lsu_response_filter.sv
```diff
@@ -0,0 +1,151 @@
...
+// is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS
+// OF ANY KIND, EITHER EXPRESSED OR IMPLIED.
+//
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+////////////////////////////////////////////////////////////////////////////////
+// Engineer:       Halfdan Bechmann  -  halfdan.bechmann@silabs.com           //
+//                                                                            //
+// Design Name:    Write Buffer                                               //
```
Silabs-ArjanB, line 24: Fix design name

rtl/cv32e40x_lsu_response_filter.sv
```diff
@@ -0,0 +1,151 @@
...
+
+  logic                  trans_accepted;
+
+  logic                  bufferable_resp; // Bufferable transfer response
+  logic                  bufferable_resp_q;
+  logic                  non_bufferable_pending;
+  logic                  resp_is_bufferable;
+
+  // Shift register containing bufferable cofiguration of outstanding transfers
+  logic [DEPTH:0]        outstanding_bufferable_q;
```
silabs-oysteink, line 71: Is the shift register deliberately DEPTH+1 bits?

rtl/cv32e40x_lsu_response_filter.sv
```diff
@@ -0,0 +1,151 @@
...
+  // The transfer accepted pulse is therefore extended until the pending non-bufferable transfer is finished.
+  assign bufferable_resp =  trans_accepted || (bufferable_resp_q && non_bufferable_pending);
+
+  always_comb begin
+    outstanding_bufferable_next = outstanding_bufferable_q;
+
+    if (trans_accepted) begin
+      // Shift in bufferable bit of accepted transfer
+      outstanding_bufferable_next    = outstanding_bufferable_next << 1;
+      outstanding_bufferable_next[1] = trans_i.memtype[0];
```
silabs-oysteink, line 91: Index 0 seems unused, if on purpose it should be explained in a comment.

rtl/cv32e40x_load_store_unit.sv
```diff
@@ -424,7 +431,7 @@ module cv32e40x_load_store_unit import cv32e40x_pkg::*;
                                                           id_ex_pipe_i.alu_operand_a;
 
   // Busy if there are ongoing (or potentially outstanding) transfers
-  assign busy_o = (cnt_q != 2'b00) || trans_valid;
+  assign busy_o = resp_filter_busy || trans_valid;
```
Silabs-ArjanB, line 434: Can you add an assertion that cnt_q != 2'b00 implies resp_filter_busy?

rtl/cv32e40x_load_store_unit.sv
```diff
@@ -126,6 +131,8 @@ module cv32e40x_load_store_unit import cv32e40x_pkg::*;
   logic         misaligned_halfword;     // high if a halfword is not naturally aligned but no split is needed
   logic         misaligned_access;       // Access is not naturally aligned
 
+  logic         resp_filter_busy;        // Response filter busy
```
Silabs-ArjanB, line 134: Rename to filter_resp_busy

silabs-oysteink (changes requested):

Silabs-ArjanB (changes requested):
Just requesting some more comments and some signal renaming

Silabs-ArjanB (changes requested):
Also please add a subsection in the LSU section of the user manual explaining that there is one entry deep write buffer and explain which transactions can(not) get buffered.

rtl/cv32e40x_lsu_response_filter.sv
```diff
@@ -0,0 +1,151 @@
...
+
+  // Blocking transfers when outstanding queue is full to avoid cnt_q overflow
+  assign ready_o      = ready_i && (cnt_q < DEPTH);
+  assign valid_o      = valid_i && (cnt_q < DEPTH);
+
+  // Response Mux
+  // Signals resp_valid early for bufferable transfers and lets resp_valid through when waiting for a non-bufferable transfers
+  // The early response for bufferable transfers is gated if we already have an outstanding bufferable transfer to ensure
+  // Correct ordering of the responses
+  assign resp_valid_o = (resp_is_bufferable) ? bufferable_resp_q && !non_bufferable_pending : resp_valid_i;
```
Silabs-ArjanB, line 147: If resp_is_bufferable = 1, why does it matter at all what other transactions are in the FIFO? For example if the oldest transaction in the FIFO is bufferable and it is followed by a non-bufferable transaction in the FIFO, I would still expect that this oldest transaction gets its resp_valid_o = 1 immediately (no matter what else is in the FIFO). Of course there needs to be RTL that takes care that each transaction in the FIFO results in exactly 1 resp_valid_o = 1. I would expect two counters: One that points to the oldest transaction that still needs to receive an resp_valid_i and one that points to the oldest transaction for which resp_valid_o is being generated.

rtl/cv32e40x_lsu_response_filter.sv
silabs-halfdan: Yes, will add comment.

rtl/cv32e40x_lsu_response_filter.sv
silabs-halfdan: Correct, index 0 is unused, done to match cnt_q values directly. Will add comment.

rtl/cv32e40x_load_store_unit.sv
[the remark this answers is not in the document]
silabs-halfdan: Will fix indentation and make explicit signal past the write buffer.

rtl/cv32e40x_load_store_unit.sv
silabs-halfdan: Good point, theses should not be mixed. Will discuss the todo w/ Øystein.

rtl/cv32e40x_lsu_response_filter.sv
silabs-halfdan: Will do...

rtl/cv32e40x_load_store_unit.sv
silabs-halfdan: Yes, that would be a good assertion. Will add.

rtl/cv32e40x_load_store_unit.sv
silabs-halfdan: Yes, that's better.

rtl/cv32e40x_lsu_response_filter.sv
silabs-halfdan: This is a good point. This is a limitation and it is not needed. Adding a counter on the core side of the response filter will (or use the already implemented lsu counter) will remove the limitation and make the code cleaner.
I will update the design and documentation to use two counters instead of generating a response pulse.

rtl/cv32e40x_load_store_unit.sv
silabs-halfdan: It's only ok because the resp_err signal is unchanged by the mpu, so there is no good reason to mix prefixes here.
Will change it to use bus_resp_err instead.

Silabs-ArjanB (approved):

Silabs-ArjanB (approved):

PR #316, opened 2021-11-16
Added check CSR instruction accepted by xif AND pipeline.

silabs-oysteink:
Killing xif instruction via commit interface in EX, and marking instruction
as illegal.

Note that the commit_kill signal is not fully implemented yet. It will get updated in a future PR with proper killing for interrupts, debug and exceptions. Changes are expected to the way commit_kill is signalled, as for killed instructions ex_valid will not be 1.

### DIFF
--- a/rtl/cv32e40x_ex_stage.sv
+++ b/rtl/cv32e40x_ex_stage.sv
@@ -55,6 +55,9 @@ module cv32e40x_ex_stage import cv32e40x_pkg::*;
   output logic        branch_decision_o,
   output logic [31:0] branch_target_o,
 
+  // Output to controller
+  output logic        xif_csr_error_o,
+
   // LSU handshake interface
   input  logic        lsu_valid_i,
   output logic        lsu_ready_o,
@@ -114,6 +117,10 @@ module cv32e40x_ex_stage import cv32e40x_pkg::*;
   assign lsu_en_gated = id_ex_pipe_i.lsu_en && instr_valid; // Factoring in instr_valid to suppress bus transactions on kill/halt
 
   assign div_en = id_ex_pipe_i.div_en && id_ex_pipe_i.instr_valid; // Valid DIV in EX, not affected by kill/halt
+
+  // If pipeline is handling a valid CSR AND the same instruction is accepted by the eXtension interface
+  // we need to convert the instruction to an illegal instruction.
+  assign xif_csr_error_o = id_ex_pipe_i.xif_en && (id_ex_pipe_i.csr_en && !csr_illegal_i);
 
 
   // Exception happened during IF or ID, or trigger match in ID (converted to NOP).
@@ -288,8 +295,10 @@ module cv32e40x_ex_stage import cv32e40x_pkg::*;
         ex_wb_pipe_o.instr_valid <= 1'b1;
         // Deassert rf_we in case of illegal csr instruction or
         // when the first half of a misaligned/split LSU goes to WB.
+        // Also deassert if CSR was accepted both by eXtension if and pipeline
         ex_wb_pipe_o.rf_we       <= (csr_illegal_i     ||
-                                    lsu_split_i)       ? 1'b0 : id_ex_pipe_i.rf_we;
+                                    lsu_split_i        ||
+                                    xif_csr_error_o)      ? 1'b0 : id_ex_pipe_i.rf_we;
         ex_wb_pipe_o.lsu_en      <= id_ex_pipe_i.lsu_en;
 
         if (id_ex_pipe_i.rf_we) begin
@@ -316,7 +325,7 @@ module cv32e40x_ex_stage import cv32e40x_pkg::*;
         ex_wb_pipe_o.instr_meta     <= instr_meta_n;
 
         // CSR illegal instruction detected in this stage, OR'ing in the status
-        ex_wb_pipe_o.illegal_insn   <= id_ex_pipe_i.illegal_insn || csr_illegal_i;
+        ex_wb_pipe_o.illegal_insn   <= id_ex_pipe_i.illegal_insn || csr_illegal_i || xif_csr_error_o;
 
         ex_wb_pipe_o.ebrk_insn      <= id_ex_pipe_i.ebrk_insn;
         ex_wb_pipe_o.wfi_insn       <= id_ex_pipe_i.wfi_insn;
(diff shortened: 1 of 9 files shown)


sva/cv32e40x_controller_fsm_sva.sv
```diff
@@ -276,5 +282,23 @@ module cv32e40x_controller_fsm_sva
...
+                  id_ex_pipe_i.xif_en && id_ex_pipe_i.csr_en && !csr_illegal_i
+                  |-> xif_commit_kill)
+    else `uvm_error("controller", "Duplicate CSR instruction not killed")
+
+  // Assert that a CSR instruction that is accepted by both eXtension interface and pipeline
+  // causes an illegal instruction
+  a_duplicate_csr_illegal:
+    assert property (@(posedge clk) disable iff (!rst_n)
+                    ex_valid_i && wb_ready_i && id_ex_pipe_i.xif_en && id_ex_pipe_i.csr_en && !csr_illegal_i &&
+                    !((id_ex_pipe_i.instr.mpu_status != MPU_OK) || (id_ex_pipe_i.instr.bus_resp.err))
```
Silabs-ArjanB, line 298: Can you add a todo here? The '!((id_ex_pipe_i.instr.mpu_status != MPU_OK) || (id_ex_pipe_i.instr.bus_resp.err))' part should not be needed once instruction offload is implemented according to spec (forbidding offload exactly for these conditions and more)

rtl/cv32e40x_ex_stage.sv
```diff
@@ -316,7 +325,7 @@ module cv32e40x_ex_stage import cv32e40x_pkg::*;
         ex_wb_pipe_o.instr_meta     <= instr_meta_n;
 
         // CSR illegal instruction detected in this stage, OR'ing in the status
-        ex_wb_pipe_o.illegal_insn   <= id_ex_pipe_i.illegal_insn || csr_illegal_i;
+        ex_wb_pipe_o.illegal_insn   <= id_ex_pipe_i.illegal_insn || csr_illegal_i || xif_csr_error_o;
```
Silabs-ArjanB, line 328: I think all (or at least most) uses of csr_illegal_i are now not correct. Consider the case where a CSR instruction is offloaded and accepted by a coprocessor and leads to csr_illegal_i in the core. In that case no illegal_insn should be caused. So basically you get an illegal instruction here if neither the coproc nor the processor accepts the CSR instruction or if both accept it. If exactly one of them accepts it, then there is no issue.

rtl/cv32e40x_ex_stage.sv
```diff
@@ -115,6 +118,10 @@ module cv32e40x_ex_stage import cv32e40x_pkg::*;
 
   assign div_en = id_ex_pipe_i.div_en && id_ex_pipe_i.instr_valid; // Valid DIV in EX, not affected by kill/halt
 
+  // If pipeline is handling a valid CSR AND the same instruction is accepted by the eXtension interface
+  // we need to convert the instruction to an illegal instruction.
+  assign xif_csr_error_o = id_ex_pipe_i.xif_en && (id_ex_pipe_i.csr_en && !csr_illegal_i);
```
Silabs-ArjanB, line 123: I am always a bit paranoid when instr_valid is not used. Can you explain why that signal is not needed here?

sva/cv32e40x_controller_fsm_sva.sv
silabs-oysteink: Good point, I'll add a todo.

rtl/cv32e40x_ex_stage.sv
silabs-oysteink: You're right, this need to change.

rtl/cv32e40x_ex_stage.sv
```diff
@@ -309,7 +318,8 @@ module cv32e40x_ex_stage import cv32e40x_pkg::*;
         end
 
         // Update signals for CSR access in WB
-        // deassert csr_en in case of illegal csr instruction
+        // deassert csr_en in case of an internal illegal csr instruction
+        // to avoid writing to CSRs inside the core.
```
Silabs-ArjanB, line 322: On the next line I think that csr_en needs to be suppressed based on csr_is_illegal as well

rtl/cv32e40x_ex_stage.sv
```diff
@@ -119,9 +122,16 @@ module cv32e40x_ex_stage import cv32e40x_pkg::*;
   assign div_en = id_ex_pipe_i.div_en && id_ex_pipe_i.instr_valid; // Valid DIV in EX, not affected by kill/halt
 
   // If pipeline is handling a valid CSR AND the same instruction is accepted by the eXtension interface
-  // we need to convert the instruction to an illegal instruction.
-  assign xif_csr_error_o = id_ex_pipe_i.xif_en && (id_ex_pipe_i.csr_en && !csr_illegal_i);
+  // we need to convert the instruction to an illegal instruction and signal commit_kill to the eXtension interface.
+  // Using registered instr_valid, as gated instr_valid would not allow killing of offloaded instruction
+  // in case of halt_ex==1. We need to kill this duplicate regardless of halt state.
```
Silabs-ArjanB, line 127: I had specified that we would kill or commit offloaded instruction specifically only when halt_ex = 0. Are you saying that there are scenarios where we cannot wait for that to happen?

Silabs-ArjanB (changes requested):

rtl/cv32e40x_ex_stage.sv
silabs-oysteink: An offloaded CSR instruction that is also accepted by the core (csr_illegal_i = 0) will cause csr_is_illegal to be 1. But if an offloaded CSR also causes an illegal CSR in the pipeline, csr_is_illegal would be 0, as only the XIF accepted the instruction. Thus we need to deassert csr_en by using the csr_illegal_i from cs_registers, otherwise we may end up writing random values to CSRs when the XIF should be the only one handling CSR instructions. This case is shown in formal where DCSR may be written from machine mode, unless we suppress csr_en based on csr_illegal_i.

Silabs-ArjanB: >> Thus we need to deassert csr_en by using the csr_illegal_i from cs_registers, otherwise we may end up writing random values to CSRs when the XIF should be the only one handling CSR instructions.

That is true, but there are 4 cases and not all of them are handled correctly yet:

1. CSR accepted by XIF && CSR legal in core -> ex_wb_pipe_o.csr_en does need to become suppressed to 0 (not handled in current code)
2. CSR not accepted by XIF && CSR legal in core -> ex_wb_pipe_o.csr_en does not need to be suppressed
3. CSR accepted by XIF && CSR not legal in core -> ex_wb_pipe_o.csr_en does need to become suppressed to 0 (handled in current code)
4. CSR not accepted by XIF && CSR not legal in core -> ex_wb_pipe_o.csr_en does need to become suppressed to 0 (handled in current code)

Silabs-ArjanB (approved):

PR #328, opened 2021-11-25
Adding missing zext.h instruction [B_EXT]

silabs-oysteink:
Decoding of sext.h was also fixed, as bits[31:25] did not match the specification.

### DIFF
--- a/rtl/cv32e40x_b_decoder.sv
+++ b/rtl/cv32e40x_b_decoder.sv
@@ -139,6 +139,7 @@ module cv32e40x_b_decoder import cv32e40x_pkg::*;
             end
           end
           {7'b0000100, 3'b100}: begin // Zero extend halfword (zext.h)
+            // NB: zext.h is a subset of the proposed pack instruction in Zbkb
             if (RV32B_ZBB) begin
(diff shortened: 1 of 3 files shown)


rtl/cv32e40x_b_decoder.sv
```diff
@@ -138,6 +138,14 @@ module cv32e40x_b_decoder import cv32e40x_pkg::*;
               decoder_ctrl_o.alu_operator               = ALU_B_ROR;
             end
           end
+          {7'b0000100, 3'b100}: begin // Zero extend halfword (zext.h)
```
Silabs-ArjanB, line 141: This code is not checking the 5 bits that are supposed to be 0 (and cause illegal instruction otherwise)

Silabs-ArjanB (changes requested):

rtl/cv32e40x_b_decoder.sv
silabs-oysteink: Fixed in update

silabs-halfdan (approved):

Silabs-ArjanB (approved):

PR #336, opened 2021-12-01
Changed minimum ID_WIDTH to 3

Silabs-ArjanB:


Silabs-ArjanB:
Self merge (will revert if not accepted in XIF)

PR #349, opened 2021-12-13
Fix todo's

silabs-oivind:

### DIFF
--- a/rtl/cv32e40x_core.sv
+++ b/rtl/cv32e40x_core.sv
@@ -58,7 +58,7 @@ module cv32e40x_core import cv32e40x_pkg::*;
   input  logic [31:0] dm_halt_addr_i,
   input  logic [31:0] hart_id_i,
   input  logic [31:0] dm_exception_addr_i,
-  input  logic [31:0] nmi_addr_i,               // TODO:OK:low use
+  input  logic [31:0] nmi_addr_i,
 
   // Instruction memory interface
   output logic        instr_req_o,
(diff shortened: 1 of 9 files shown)


rtl/cv32e40x_core.sv
```diff
@@ -733,7 +733,7 @@ module cv32e40x_core import cv32e40x_pkg::*;
     .irq_wu_ctrl_i                  ( irq_wu_ctrl            ),
     .irq_req_ctrl_i                 ( irq_req_ctrl           ),
     .irq_id_ctrl_i                  ( irq_id_ctrl            ),
-    .current_priv_lvl_i             ( current_priv_lvl       ), // TODO:OK:low Needs bypass for 40S?
+    .current_priv_lvl_i             ( current_priv_lvl       ),
```
Silabs-ArjanB, line 736: Can this code get aligned with that of the 40s naming-wise (so no 'current_' in the name)? (The 40S naming is better and should be used). Also please clean up code that does not actually use this signal (ID stage, decoder, int controller are the ones I noticed). Finally we need comments on 'why' it is safe to use this signal all over the place (instead of using for example a bypass)

Silabs-ArjanB (changes requested):

rtl/cv32e40x_core.sv
silabs-oivind: Everything related to priv_lvl is now removed. SEC clean.

PR #356, opened 2021-12-16
Removed todo.

silabs-oivind:
Won't do this as the 32bit 3:1 mux needed to avoid adder results in slightly larger area


Silabs-ArjanB (commented):
Did you consider re-using this mux? https://github.com/openhwgroup/cv32e40x/blob/df7aab4bb6afe105410e6befb4cfed4cb0c03320/rtl/cv32e40x_cs_registers.sv#L457

If that helps then maybe that mux should be moved into the controller fsm instead. It is hard to believe the we need yet another adder (we could also re-use the branch target adder or the ALU).

silabs-oivind:
Created PR #378 to implement what you suggested. Closing this PR.

PR #358, opened 2021-12-20
Todo removals

silabs-oysteink:
SEC clean with OBI assumes.


silabs-oivind (approved):

silabs-oysteink:
Self servicing merge due to holidays.

PR #370, opened 2022-01-05
Bug fir for atomic operations (enabled ALU as well as LSU leading to violations inside of LSU)

Silabs-ArjanB:
Bug fix; not SEC clean

Merge after https://github.com/openhwgroup/cv32e40x/pull/369


Silabs-ArjanB:
Will re-issue after  #369

PR #372, opened 2022-01-06
Fixed atomic LR decoding

Silabs-ArjanB:
Fix LR decoding (rs2 field should be checked to be 0).
Not SEC clean (SEC clean if A_EXT=0)

### DIFF
--- a/rtl/cv32e40x_a_decoder.sv
+++ b/rtl/cv32e40x_a_decoder.sv
@@ -45,10 +45,8 @@ module cv32e40x_a_decoder import cv32e40x_pkg::*;
 
       OPCODE_AMO: begin
         if (instr_rdata_i[14:12] == 3'b010) begin // RV32A Extension (word)
-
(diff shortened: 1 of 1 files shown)


rtl/cv32e40x_a_decoder.sv
```diff
@@ -58,7 +56,12 @@ module cv32e40x_a_decoder import cv32e40x_pkg::*;
 
           unique case (instr_rdata_i[31:27])
             AMO_LR: begin
-              decoder_ctrl_o.lsu_we = 1'b0;
+              decoder_ctrl_o.rf_re[1]     = 1'b0;
```
silabs-oysteink, line 59: Isn't this already 0 at this point?

silabs-oysteink (changes requested):

rtl/cv32e40x_a_decoder.sv
Silabs-ArjanB: Yes, it is, just making it explicit here

PR #373, opened 2022-01-06
Removal of todo's and currently unused logic

silabs-oysteink:
Removed logic intended for preserving address of a faulted load or store for use in mtval CSR.

This code was not used, and it would not work in the current state as it could overwrite the
faulted address.

Added assertion to check max number of retired instructions after an NMI becomes pending.
This uncovered a case where the result of a faulted load may actually be used.

Bugfix: Halting ID stage in the same cycle as the bus error arrives, allowing max 1 extra retirement
after the faulted LSU instruction..

Non-SEC due to the bugfix.

### DIFF
--- a/rtl/cv32e40x_controller_fsm.sv
+++ b/rtl/cv32e40x_controller_fsm.sv
@@ -61,7 +61,6 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
 
   // From WB stage
   input  logic [1:0]  lsu_err_wb_i,               // LSU caused bus_error in WB stage, gated with data_rvalid_i inside load_store_unit
-  input  logic [31:0] lsu_addr_wb_i,              // LSU address in WB stage //todo: only needed if MTVAL is implemented
 
   // From LSU (WB)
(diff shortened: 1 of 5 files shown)


rtl/cv32e40x_controller_fsm.sv
```diff
@@ -389,7 +398,7 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
     ctrl_fsm_o.halt_id = ctrl_byp_i.jr_stall || ctrl_byp_i.load_stall || ctrl_byp_i.csr_stall || ctrl_byp_i.wfi_stall ||
                          (pending_interrupt && !interrupt_allowed) ||
                          (pending_debug && !debug_allowed) ||
-                         (pending_nmi && !nmi_allowed);
+                         (pending_nmi_early && !nmi_allowed);
```
Silabs-ArjanB, line 401: I wonder if any of the qualification with !*allowed is needed here. Will investigate this myself after this PR.

Silabs-ArjanB (approved):

PR #378, opened 2022-01-07
Add PC mux in cv32e40x_controller_fsm. use this mux for fencei PC and exception PC

silabs-oivind:
PR consists of 3 commits:
ccd2871: SEC clean. Move PC mux from CS registers to controller FSM
f743203: Not SEC clean. Update the halt/kill behavior upon fencei. This is not SEC clean because the previous implementation would try to fetch an (unneeded) instruction when IF was killed when fencei entered WB. Halting IF (not killing) prevents this redundant instruction fetch.
5069951: SEC clean. Use output from PC mux in controller FSM upon fencei.

### DIFF
--- a/rtl/cv32e40x_controller_fsm.sv
+++ b/rtl/cv32e40x_controller_fsm.sv
@@ -44,6 +44,9 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
   // From bypass logic
   input  ctrl_byp_t   ctrl_byp_i,
 
+  // From IF stage
+  input logic [31:0]   pc_if_i,
+
   // From ID stage
   input  if_id_pipe_t if_id_pipe_i,
   input  logic        sys_mret_id_i,              // mret in ID stage
@@ -172,6 +175,9 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
   logic       fencei_flush_req_set;
   logic       fencei_req_and_ack_q;
   logic       fencei_ongoing;    
+
+  // Pipeline PC mux control
+  pipe_pc_mux_e pipe_pc_mux_ctrl;
 
   // Flag for signalling that a new instruction arrived in WB.
   // Used for performance counters. High for one cycle, unless WB is halted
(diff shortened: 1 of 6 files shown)


rtl/cv32e40x_if_stage.sv
```diff
@@ -118,7 +118,7 @@ module cv32e40x_if_stage import cv32e40x_pkg::*;
       PC_BRANCH:   branch_addr_n = branch_target_ex_i;
       PC_MRET:     branch_addr_n = mepc_i;                                                      // PC is restored when returning from IRQ/exception
       PC_DRET:     branch_addr_n = dpc_i;
-      PC_FENCEI:   branch_addr_n = ex_wb_pipe_i.pc + 4;                                         // Jump to next instruction forces prefetch buffer reload // TODO:OK:low Can avoid adder, PC should already be in pipeline
+      PC_FENCEI:   branch_addr_n = ctrl_fsm_i.pipe_pc;                                          // Jump to next instruction forces prefetch buffer reload
```
Silabs-ArjanB, line 121: ex_wb_pipe_i no longer used. Remove from interface.

Silabs-ArjanB (changes requested):

rtl/cv32e40x_cs_registers.sv
```diff
@@ -455,17 +453,6 @@ module cv32e40x_cs_registers import cv32e40x_pkg::*;
     unique case (1'b1)
 
       ctrl_fsm_i.csr_save_cause: begin
-        unique case (1'b1)
-          ctrl_fsm_i.csr_save_if:
-            exception_pc = pc_if_i;
-          ctrl_fsm_i.csr_save_id:
-            exception_pc = if_id_pipe_i.pc;
```
silabs-oysteink, line 462: if_id_pipe_i.pc is also used for the default value of the exception in this always block. Can we choose a different default and remove if_id_pipe_i from this module?

rtl/cv32e40x_controller_fsm.sv
```diff
@@ -372,6 +378,20 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
...
+  always_comb begin
+
+    ctrl_fsm_o.pipe_pc = PC_WB;
+
+    unique case (pipe_pc_mux_ctrl)
+      PC_WB: ctrl_fsm_o.pipe_pc = ex_wb_pipe_i.pc;
+      PC_EX: ctrl_fsm_o.pipe_pc = id_ex_pipe_i.pc;
+      PC_ID: ctrl_fsm_o.pipe_pc = if_id_pipe_i.pc;
+      PC_IF: ctrl_fsm_o.pipe_pc = pc_if_i;
+      default:;
```
silabs-oysteink, line 391: Is this default needed, since the mux selector uses all available entries and the case is uniqe?

silabs-oysteink (changes requested):

rtl/cv32e40x_controller_fsm.sv
Silabs-ArjanB: Please always use a default

rtl/cv32e40x_if_stage.sv
silabs-oivind: Done

rtl/cv32e40x_cs_registers.sv
silabs-oivind: Done

rtl/cv32e40x_controller_fsm.sv
silabs-oivind: Keeping default, since mux enums might change in the future.

silabs-oysteink (approved):

PR #379, opened 2022-01-07
Fixes according to renamed signals

Silabs-ArjanB:
Wrapper fixes (SEC clean)


Silabs-ArjanB:
Self service

Issue #393, opened 2022-01-14
XIF issue interface uses same instruction ID for two consecutive offloaded instructions


michael-platzer:
Component: **Component:RTL**
Git Hash: 9c1d3db

The XIF issue interface sometimes uses the same instruction ID for two consecutive offloaded instruction, although the first instruction has not retired. This happens when an instruction is accepted by the coprocessor but the pipeline is currently stalled and hence the offloaded instruction remains in the ID stage despite having been accepted. In such a situation the [`xif_accepted_q` register in the ID stage is set](https://github.com/openhwgroup/cv32e40x/blob/9c1d3db443c862d41ddf17dc0ba97ace46e78aa3/rtl/cv32e40x_id_stage.sv#L655), which [de-asserts the `issue_valid` signal](https://github.com/openhwgroup/cv32e40x/blob/9c1d3db443c862d41ddf17dc0ba97ace46e78aa3/rtl/cv32e40x_id_stage.sv#L665-L666) of the XIF issue interface (intentionally so, since the instruction has already been accepted).

However, the `issue_valid` signal is also [used by the IF stage to decide whether a new instruction ID needs to be generated](https://github.com/openhwgroup/cv32e40x/blob/9c1d3db443c862d41ddf17dc0ba97ace46e78aa3/rtl/cv32e40x_if_stage.sv#L292). Hence, if the ID stage is stalled while an offloaded instruction has already been accepted, then the IF stage does not increment the instruction ID. This leads to the same ID being assigned to the next instruction.

The problem is shown in the waveforms below. Two instruction with IDs 0 and 1 are offloaded normally, as shown by the signals `issue_valid` and `issue_ready` (with the associated ID shown by `issue_id`). The instruction ID in the IF stage (signal `xif_id`) is incremented for each of these two instructions. Next, an instruction with ID 2 is offloaded and accepted by the coprocessor. However, the ID stage is stalled and thus `xif_accepted_q` is set to indicate that the instruction in the ID stage has already been accepted. While `xif_id` briefly switches to 3 while the coprocessor accepts instruction 2, it then goes back to 2 once `issue_valid` is no longer asserted. Consequently, when the next instruction enters the ID stage, the associated instruction ID is still 2 and the XIF interface re-uses the same ID, although the previous instruction has not retired.

![Screenshot from 2022-01-14 17-24-49](https://user-images.githubusercontent.com/9446837/149551897-2fc072d1-54f7-4dbc-b43e-f64fb71e7e9a.png)

PR #394, opened 2022-01-14
Fix XIF ID increment logic (#393)

michael-platzer:
This is a patch for issue #393.

The IF stage increments XIF instruction IDs when the `issue_valid`
signal of the XIF issue interface is asserted.  However, if the ID stage
is stalled but the instruction currently held in it has been offloaded
and already accepted by the coprocessor, then `issue_valid` signal is no
longer asserted and thus the same instruction ID is reused for the next
offloaded instruction.

This patch fixes the problem by introducing a new signal that indicates
whether the instruction currently held in the ID stage is being
offloaded, with that signal remaining asserted if the coprocessor has
already accepted an instruction. The new signal informs the IF stage
whether the instruction currently in ID is being offloaded, in which
case the instruction ID is incremented.

### DIFF
--- a/rtl/cv32e40x_core.sv
+++ b/rtl/cv32e40x_core.sv
@@ -235,7 +235,10 @@ module cv32e40x_core import cv32e40x_pkg::*;
   logic        irq_ack;
   logic [4:0]  irq_id;
   logic        dbg_ack;
-  
+
+  // eXtension interface signals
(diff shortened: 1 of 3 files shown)


Silabs-ArjanB (approved):
Some minor changes to comply with our naming style

Issue #395, opened 2022-01-17
tcontrol CSR present in UM but not in RTL


silabs-hfegran:
## `tcontrol` CSR (0x7A5) is present in the user manual but not in RTL

Present in UM but not in RTL, any access attempts traps.

![image](https://user-images.githubusercontent.com/75469886/149753488-bbe411f3-de2a-4888-81a4-37c0702ca1d4.png)

PR #397, opened 2022-01-17
Added mising CSR tcontrol at address 0x7A5.

silabs-oysteink:
Always reads zero, writes are ignored.
Should fix issue #395.


Silabs-ArjanB (approved):
Will need to take DBG_NUM_TRIGGERS into account later as well.

PR #401, opened 2022-01-20
Add M_EXT parameter

silabs-oivind:
Add M_EXT parameter.
Update MISA to take M_EXT, X_EXT, X_MISA into account

### DIFF
--- a/rtl/cv32e40x_m_decoder.sv
+++ b/rtl/cv32e40x_m_decoder.sv
@@ -27,6 +27,9 @@
 ////////////////////////////////////////////////////////////////////////////////
 
 module cv32e40x_m_decoder import cv32e40x_pkg::*;
+  #(
+    parameter m_ext_e M_EXT = M
+    )
   (
    // from IF/ID pipeline
    input logic [31:0] instr_rdata_i,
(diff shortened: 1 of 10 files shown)


rtl/cv32e40x_m_decoder.sv
```diff
@@ -77,24 +80,32 @@ module cv32e40x_m_decoder import cv32e40x_pkg::*;
             decoder_ctrl_o.mul_operator     = MUL_H;
           end
           {7'b000_0001, 3'b100}: begin // div
-            decoder_ctrl_o.div_en           = 1'b1;
-            decoder_ctrl_o.div_operator     = DIV_DIV;
-            decoder_ctrl_o.alu_operator     = ALU_SLL;
+            if (M_EXT == M) begin
```
Silabs-ArjanB, line 83: All of these need an 'else' branch to cause an illegal instruction. I prefer if you would simply set the following in that branch:

decoder_ctrl_o = DECODER_CTRL_ILLEGAL_INSN;

Silabs-ArjanB (changes requested):
Apart from the required decoder fixes we also need generate statements around multi_i and div_i.

rtl/cv32e40x_m_decoder.sv
silabs-oivind: Yes, of course.. Added.

silabs-oivind:
> Apart from the required decoder fixes we also need generate statements around multi_i and div_i.

Added generate statements. Synth tools should be able to optimize away these instances without the generate statement because their outputs are static/not used. But agree it's cleaner to add generate statements.

Silabs-ArjanB (approved):

Issue #398, opened 2022-01-18
Assert a_nmi_handler_max_retire fails in formal with PMA config


silabs-mateilga:
**Component:RTL**
git hash: 9c1d3db443c862d41ddf17dc0ba97ace46e78aa3

Controller retires two instructions (1 is max) before taking the bus error NMI, when an ST instruction is retired to the write buffer. See waveform0. for details.

Early debugging with @silabs-oysteink indicates id stage should be halted in the case where nmi_allowed is high but the fsm decides not to take it anyway. 
A quick modification where nmi_allowed is made to depend on nmi_pending_q (   assign nmi_allowed = interrupt_allowed && nmi_pending_q;   ) solved the immediate CEX, but formal found another CEX seen in waveform1.

reproduce: Run formal verification with PMA enabled, any of the PMA configurations found in core-v-verif/cv32e40x/tb/uvmt/uvmt_cv32e40x_constants.sv should show the problem, example waveform from pma_test_cfg_4.

waveform0
![nmiAssert](https://user-images.githubusercontent.com/59358563/149933598-63e20025-db45-43d8-a5f1-a704ca1a3823.png)

waveform1
![nmiAssertModified](https://user-images.githubusercontent.com/59358563/149935174-5695495d-0c1c-4ccb-86ec-83bfd7c31e84.png)

Silabs-ArjanB:
@silabs-mateilga Can this be closed now?

silabs-mateilga:
@Silabs-ArjanB yes, closing it now.

PR #405, opened 2022-01-24
Fix for issue #398.

silabs-oysteink:
Three instructions could be retired between a bus error and the taken NMI.
Fixed RTL to halt ID stage earlier, and updated the associated assertion
to take into account that bus errors for bufferable writes may come between
instructions.


silabs-oysteink:
Please wait with the merge, I need to investigate a couple more scenarios.

Silabs-ArjanB (approved):

Issue #156, opened 2021-07-28
csr_en not suppressed for killed instructions


Silabs-ArjanB:
In the ID stage csr_en is not suppressed for nullified instructions. This leads to the possibility that csr_en in ID is 1 while alu_en in ID is 0, such that the ALU operands into EX will not get updated (these are also used for CSR instructions). This doesn't cause a functional issue, but it causes different stall cycles and therefore SEC issues. It would be best to suppress csr_en (similar to how alu_en is suppressed) (a first attempt however led to timing issues).

The following line https://github.com/openhwgroup/cv32e40x/blob/e4b1032774b5269f93e0b6912ff1d6cf11b2ef24/rtl/cv32e40x_id_stage.sv#L529 is temporarily changed in https://github.com/openhwgroup/cv32e40x/pull/155. The change from that PR should be undone together with this fix. It should also be checked that replacing 'if (alu_en || div_en)' by 'if (1)' is SEC equivalent.

PR #411, opened 2022-01-28
Suppressing csr_en for nullified instructions 

silabs-halfdan:
Suppressing csr_en for nullified instructions instead of csr_op to avoid affecting stall cycles with nullified instructions.
This no longer results in timing issues.

Fix for #156

### DIFF
--- a/rtl/cv32e40x_decoder.sv
+++ b/rtl/cv32e40x_decoder.sv
@@ -98,6 +98,7 @@ module cv32e40x_decoder import cv32e40x_pkg::*;
 
   csr_opcode_e csr_op;
 
+  logic       csr_en;
   logic       alu_en;
   logic       mul_en;
   logic       div_en;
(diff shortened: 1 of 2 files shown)


silabs-oysteink (approved):

rtl/cv32e40x_decoder.sv
```diff
@@ -229,7 +228,7 @@ module cv32e40x_decoder import cv32e40x_pkg::*;
   assign div_en_o = deassert_we_i ? 1'b0        : div_en;
   assign lsu_en_o = deassert_we_i ? 1'b0        : lsu_en;
 
-  assign csr_op_o = deassert_we_i ? CSR_OP_READ : csr_op; // Gating csr_en introduces tight timing path, gate csr_op instead
+  assign csr_en_o = deassert_we_i ? 1'b0        : csr_en;
   assign rf_we_o  = deassert_we_i ? 1'b0        : rf_we;
```
Silabs-ArjanB, line 232: Can id_ex_pipe_o.csr_op                 <= csr_op; in the ID stage now be gated with the related enable signal?

silabs-halfdan: Yes, that should be ok (and align with the other ID enables). I'll check with SEC and push the change.

Silabs-ArjanB (approved):

PR #414, opened 2022-02-02
Updated signal name in sva

silabs-hfegran:
Fixed signal name update that was missed in previous commit

Signed-off-by: Henrik Fegran <Henrik.Fegran@silabs.com>


Silabs-ArjanB (approved):
Sorry for that!

PR #421, opened 2022-02-09
Changed RO into R to reduce confusion

Silabs-ArjanB:
@silabs-oivind This commit should not be merged into the 40S repos (I will do a separate but similar commit there).


silabs-robin (approved):
This looks as expected from our discussions.

docs/user_manual/source/control_status_registers.rst
```diff
@@ -1464,7 +1462,7 @@ Detailed:
 
   CSR Address: 0x747
 
-  Reset Value: 0x0000_0000
+  Reset Value: defined (based on ``PMP_MSECCFG_RV``)
```
silabs-robin, line 1465: Wait a second... Isn't this 40s-specific?

silabs-robin: Is `.. only:: PMP` some "feature flag" that allows 40x/40s to share some documentation source but only be visible if the flag is enabled?

Silabs-ArjanB: Yes, correct. We try to use the same rst files as much as possible, but we have some conditional rendering. (It is not applied to the Xsecure CSRs, but I can introduce such a flag later as well to further reduce the diff between the two repos.

PR #440, opened 2022-02-16
Add XIF memory interface

michael-platzer:
This PR adds the XIF memory interfaces (i.e., the [memory request/response](https://docs.openhwgroup.org/projects/openhw-group-core-v-xif/x_ext.html#memory-request-response-interface) and the [memory result](https://docs.openhwgroup.org/projects/openhw-group-core-v-xif/x_ext.html#memory-result-interface) interfaces) to CV32E40X.

The first commit replaces the [`size` field in the XIF interface definitions](https://github.com/openhwgroup/cv32e40x/blob/380b9fd6dff586d332886e0eaa875e9aafa54ca6/rtl/if_xif.sv#L86) with a byte enable mask as discussed here: https://github.com/openhwgroup/core-v-xif/issues/17

The second commit implements the base functionality required to handle memory requests by a coprocessor. Memory requests received via the XIF memory interface are converted into memory transactions. The aligner performs alignment for XIF memory data based on the lower bits of the address as usual, with the assumption that an XIF request always uses the entire width of the memory interface (i.e., is always 32 bits wide). Note that the byte enable mask received via the XIF interface is aligned as well. Split accesses are performed if required (as is done for internal memory transactions).

The third commit forwards MPU errors and bus errors resulting from XIF memory transactions to the coprocessor. MPU errors must be reported immediately via the XIF memory response interface, rather than being buffered as is done for internal transactions. This required some changes to the MPU, which has now the option to issue MPU errors via the new `core_mpu_err_o` port immediately.

As with previous patches related to the XIF interface, the new functionality is only enabled when `X_EXT` is 1. In particular, if `X_EXT` is 0, then the functionality of CV32E40X is not modified.

Acknowledgements: The changes in this PR have been inspired by earlier work by Kunal Dalal, Oana Lazar, Yu Xia, and Adomas Lebedys from the University of Southampton in the context of the AVA vector accelerator project, see: https://github.com/AI-Vector-Accelerator

### DIFF
--- a/rtl/cv32e40x_load_store_unit.sv
+++ b/rtl/cv32e40x_load_store_unit.sv
@@ -28,6 +28,8 @@
 module cv32e40x_load_store_unit import cv32e40x_pkg::*;
 #(
   parameter bit          A_EXT = 0,
+  parameter bit          X_EXT = 0,
+  parameter int          X_ID_WIDTH = 4,
   parameter int          PMA_NUM_REGIONS = 0,
   parameter pma_region_t PMA_CFG[PMA_NUM_REGIONS-1:0] = '{default:PMA_R_DEFAULT}
 )
@@ -139,17 +141,38 @@ module cv32e40x_load_store_unit import cv32e40x_pkg::*;
 
   logic           trans_valid_q;        // trans_valid got clocked without trans_ready
 
+  logic                  xif_req;       // The ongoing memory request comes from the XIF interface
+  logic                  xif_ready_1;   // The LSU second stage is ready for an XIF transaction
+  logic                  xif_res_q;     // The next memory result is for the XIF interface
+  logic [X_ID_WIDTH-1:0] xif_id_q;      // Instruction ID of an XIF memory transaction
+
+  assign xif_req = X_EXT & xif_mem_if.mem_valid;
+
   // Transaction (before aligner)
   // Generate address from operands (atomic memory transactions do not use an address offset computation)
-  assign trans.addr  = id_ex_pipe_i.lsu_atop[5] ? id_ex_pipe_i.alu_operand_a : (id_ex_pipe_i.alu_operand_a + id_ex_pipe_i.alu_operand_b);
-  assign trans.we    = id_ex_pipe_i.lsu_we;
-  assign trans.size  = id_ex_pipe_i.lsu_size;
-  assign trans.wdata = id_ex_pipe_i.operand_c;
-  assign trans.mode  = PRIV_LVL_M; // Machine mode
-  assign trans.dbg   = ctrl_fsm_i.debug_mode;
+  always_comb begin
+    if (xif_req) begin
+      trans.addr  = xif_mem_if.mem_req.addr;
+      trans.we    = xif_mem_if.mem_req.we;
+      trans.size  = 2'b10;                    // XIF memory requests always use maximum size
+      trans.wdata = xif_mem_if.mem_req.wdata;
+      trans.mode  = xif_mem_if.mem_req.mode;  // TODO use mode from XIF request or force machine mode?
+      trans.dbg   = '0;                       // TODO setup debug triggers
 
-  assign trans.atop  = id_ex_pipe_i.lsu_atop;
-  assign trans.sext  = id_ex_pipe_i.lsu_sext;
+      trans.atop  = '0;
+      trans.sext  = '0;
+    end else begin
+      trans.addr  = id_ex_pipe_i.lsu_atop[5] ? id_ex_pipe_i.alu_operand_a : (id_ex_pipe_i.alu_operand_a + id_ex_pipe_i.alu_operand_b);
+      trans.we    = id_ex_pipe_i.lsu_we;
+      trans.size  = id_ex_pipe_i.lsu_size;
+      trans.wdata = id_ex_pipe_i.operand_c;
+      trans.mode  = PRIV_LVL_M; // Machine mode
+      trans.dbg   = ctrl_fsm_i.debug_mode;
+
+      trans.atop  = id_ex_pipe_i.lsu_atop;
+      trans.sext  = id_ex_pipe_i.lsu_sext;
+    end
+  end
 
   ///////////////////////////////// BE generation ////////////////////////////////
   always_comb
@@ -180,23 +203,23 @@ module cv32e40x_load_store_unit import cv32e40x_pkg::*;
         end
       end
       default:
-      begin // Writing a word
+      begin // Writing a word (note: XIF memory requests always write a word)
         if (split_q == 1'b0)
         begin // non-misaligned case
           case (trans.addr[1:0])
-            2'b00: be = 4'b1111;
-            2'b01: be = 4'b1110;
-            2'b10: be = 4'b1100;
-            2'b11: be = 4'b1000;
+            2'b00: be = xif_req ?  xif_mem_if.mem_req.be[3:0]        : 4'b1111;
+            2'b01: be = xif_req ? {xif_mem_if.mem_req.be[2:0], 1'b0} : 4'b1110;
+            2'b10: be = xif_req ? {xif_mem_if.mem_req.be[1:0], 2'b0} : 4'b1100;
+            2'b11: be = xif_req ? {xif_mem_if.mem_req.be[0:0], 3'b0} : 4'b1000;
           endcase; // case (trans.addr[1:0])
         end
         else
         begin // misaligned case
           case (trans.addr[1:0])
             2'b00: be = 4'b0000; // this is not used, but included for completeness
-            2'b01: be = 4'b0001;
-            2'b10: be = 4'b0011;
-            2'b11: be = 4'b0111;
+            2'b01: be = xif_req ? {3'b0, xif_mem_if.mem_req.be[3:3]} : 4'b0001;
+            2'b10: be = xif_req ? {2'b0, xif_mem_if.mem_req.be[3:2]} : 4'b0011;
+            2'b11: be = xif_req ? {1'b0, xif_mem_if.mem_req.be[3:1]} : 4'b0111;
           endcase; // case (trans.addr[1:0])
         end
       end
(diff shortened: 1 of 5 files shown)


Silabs-ArjanB:
Hello @michael-platzer , thank you again for your great contributions! We will review it and then merge it. In the mean time could you please have a look at the failing Eclipse ECA check; did ou maybe not sign the commit or use a different github account?

Silabs-ArjanB:
@michael-platzer Thanks!

rtl/cv32e40x_load_store_unit.sv
```diff
@@ -139,17 +141,39 @@ module cv32e40x_load_store_unit import cv32e40x_pkg::*;
...
-  assign trans.we    = id_ex_pipe_i.lsu_we;
-  assign trans.size  = id_ex_pipe_i.lsu_size;
-  assign trans.wdata = id_ex_pipe_i.operand_c;
-  assign trans.mode  = PRIV_LVL_M; // Machine mode
-  assign trans.dbg   = ctrl_fsm_i.debug_mode;
-
-  assign trans.atop  = id_ex_pipe_i.lsu_atop;
-  assign trans.sext  = id_ex_pipe_i.lsu_sext;
+  always_comb begin
+    if (xif_req) begin
```
silabs-robin, line 155: Is this conditional always used, regardless of X_EXT param?
Should it be a generate block, or have an assert that `!X_EXT |-> !xif_req`?

michael-platzer: Yes, this conditional is always used, but the assignment of `xif_req` is guarded by `X_EXT`:

https://github.com/openhwgroup/cv32e40x/blob/3aee97ac2e58b226e4184c57318ac4ef72900cc4/rtl/cv32e40x_load_store_unit.sv#L150

I you prefer, I can add the `X_EXT` parameter to all conditionals involving `xif_req` instead, although that might clutter the code. An assertion might be a good idea though, as well as for `xif_res_q`.

michael-platzer: @silabs-robin I have tentatively added assertions for the `xif_req` and `xif_res_q` signals in the corresponding `*_sva.sv` file. Please let me know if this is correct and whether I have overlooked something, since I have no experience with verification.

rtl/cv32e40x_load_store_unit.sv
```diff
@@ -139,17 +141,39 @@ module cv32e40x_load_store_unit import cv32e40x_pkg::*;
 
   logic           trans_valid_q;        // trans_valid got clocked without trans_ready
 
+  logic                  xif_req;       // The ongoing memory request comes from the XIF interface
+  logic                  xif_mpu_err;   // The ongoing memory request caused an MPU error
+  logic                  xif_ready_1;   // The LSU second stage is ready for an XIF transaction
+  logic                  xif_res_q;     // The next memory result is for the XIF interface
+  logic [X_ID_WIDTH-1:0] xif_id_q;      // Instruction ID of an XIF memory transaction
+
+  assign xif_req = X_EXT & xif_mem_if.mem_valid;
```
Silabs-ArjanB, line 150: & -> && (we use && for logical and)

rtl/cv32e40x_load_store_unit.sv
```diff
@@ -248,14 +272,24 @@ module cv32e40x_load_store_unit import cv32e40x_pkg::*;
...
-      if(!valid_0_i) begin
+      if(!valid_0_i && !xif_req) begin
         split_q <= 1'b0; // Reset split_st when no valid instructions
       end else if (ctrl_update) begin // EX done, update split_q for next address phase
         split_q <= lsu_split_0_o;
       end
     end
   end
 
+  always_ff @(posedge clk, negedge rst_n) begin
```
Silabs-ArjanB, line 283: Can you use a generate block here based on X_EXT and set the _q values to 0 in the else part?

rtl/cv32e40x_load_store_unit.sv
```diff
@@ -451,16 +485,19 @@ module cv32e40x_load_store_unit import cv32e40x_pkg::*;
...
-  assign ready_1_o = ((cnt_q == 2'b00) ? 1'b1 : resp_valid) && ready_1_i;
+  // XIF transactions bypass the pipeline, hence ready_1_i is not required for the second stage to
+  // be ready for XIF transactions.
+  assign ready_1_o   = ((cnt_q == 2'b00) ? 1'b1 : resp_valid) && ready_1_i;
+  assign xif_ready_1 = ((cnt_q == 2'b00) ? 1'b1 : resp_valid);
 
   // LSU second stage is valid when resp_valid (typically data_rvalid_i) is received. For a misaligned/split
   // load/store only its second phase is marked as valid (last_q == 1'b1)
-  assign valid_1_o = last_q && resp_valid && valid_1_i;
+  assign valid_1_o                          = last_q && resp_valid && valid_1_i && ~xif_res_q;
```
Silabs-ArjanB, line 499: ~xif_res_q -> !xif_res_q (we use ! for single bit inversions)

rtl/cv32e40x_load_store_unit.sv
```diff
@@ -474,31 +511,36 @@ module cv32e40x_load_store_unit import cv32e40x_pkg::*;
...
+                     ) && (ready_0_i || (xif_req && xif_ready_1));
 
+  // XIF request bypasses pipeline (does not assert valid_0_o) and takes precedence over regular
+  // requests, hence inhibits a concurrent request from EX stage
   assign valid_0_o =  (
                        (cnt_q == 2'b00) ? (trans_valid && trans_ready) :
                        (cnt_q == 2'b01) ? (trans_valid && trans_ready) :
                        1'b1
-                      ) && valid_0_i;
+                      ) && valid_0_i && !xif_req;
```
Silabs-ArjanB, line 529: Can you simply add a 'todo' here as this seems to be the only place where valid_o_i could not be replaced by valid_0_i || xif_req. 

Can valid_0_i and xif_req both be 1 at the same time in the current design? If not, then it would be better to remove the && !xif_req part and add an assertion instead that both bits cannot be 1 at the same time. If they can be 1 at the same time can you please add a comment on for what scenario that could happen and why xif should take precedence?

rtl/cv32e40x_load_store_unit.sv
```diff
@@ -615,6 +657,8 @@ module cv32e40x_load_store_unit import cv32e40x_pkg::*;
     .misaligned_access_i  ( misaligned_access  ),
 
     .core_one_txn_pend_n  ( cnt_is_one_next    ),
+    .core_mpu_err_wait_i  ( ~xif_req           ),
```
Silabs-ArjanB, line 660: ~xif_req -> !xif_req

rtl/cv32e40x_load_store_unit.sv
```diff
@@ -701,10 +745,21 @@ module cv32e40x_load_store_unit import cv32e40x_pkg::*;
     .m_c_obi_data_if    ( m_c_obi_data_if )
   );
 
-  // Drive eXtension interface outputs to 0 for now
-  assign xif_mem_if.mem_ready               = '0;
-  assign xif_mem_if.mem_resp                = '0;
-  assign xif_mem_result_if.mem_result_valid = '0;
-  assign xif_mem_result_if.mem_result       = '0;
+  //////////////////////////////////////////////////////////////////////////////
```
Silabs-ArjanB, line 748: Can you use a generate statement here based on X_EXT?

Silabs-ArjanB (changes requested):
Hi @michael-platzer Thanks for this PR and apologies for the slow feedback. I am just requesting some minor changes here. I will also ask Øivind to double check the MPU related changes (but those look good to me). After these minor changes I will merge this. At the same time I will send you an email with respect to an alternative proposal related to the 'be'/'size' discussion (we can make change to the RTL later if you agree with that proposal).

rtl/cv32e40x_load_store_unit.sv
michael-platzer: Yes, `valid_0_i` and `xif_req` can both be 1 at the same time when an XIF load/store instruction is immediately followed by a regular load/store instruction: In that situation, simultaneous assertion of `valid_0_i` and `xif_req` happens when the XIF load/store awaits a result in the WB stage, the regular load/store has moved to the EX stage, and the coprocessor starts the memory transaction. Loads/stores should happen in program order, hence the XIF transaction needs to take precedence.

Note that simultaneous assertion of `valid_0_i` and `xif_req` should not happen in the reverse situation (regular load/store instruction immediately followed by an XIF load/store instruction), since CV32E40X needs to wait for the memory response of the regular load/store before committing the XIF load/store (and the coprocessor must not start the XIF memory request until the instruction has been committed).

I can either add a comment explaining that situation or a TODO, please let me know which one you prefer. Do you have some prospective task in mind that I should add to the TODO (such as merging the `valid_0_i` and `xif_req` signals)?

michael-platzer:
Thanks @Silabs-ArjanB for your feedback! I have applied the requested changes related to the logical operators and the generate blocks. Please see my comment above regarding the simultaneous assertion of `valid_0_i` and `xif_req`.

I am looking forward to hearing about your alternative proposal w.r.t the `size` vs `be` field in the XIF memory request.

rtl/cv32e40x_load_store_unit.sv
Silabs-ArjanB: Hi @michael-platzer, thanks. I will just add above info into a new github ticket related to the introduction of a XIF related scoreboard. I expect that eventually we will make sure that valid-0_i and xif_req cannot be 1 at the same time. For now it is good enough to understand that it can actually happen.

Silabs-ArjanB (approved):

silabs-oivind:
> Hi @michael-platzer Thanks for this PR and apologies for the slow feedback. I am just requesting some minor changes here. I will also ask Øivind to double check the MPU related changes (but those look good to me). After these minor changes I will merge this. At the same time I will send you an email with respect to an alternative proposal related to the 'be'/'size' discussion (we can make change to the RTL later if you agree with that proposal).

Hi @michael-platzer. The MPU changes look good to me. But the MPU is instantiated in both the LSU and the IF stage. You need to update the MPU interface for the new signals in cv32e40x_if_stage as well.

michael-platzer:
@silabs-oivind Apologies for overlooking the second MPU in the IF stage and thanks @Silabs-ArjanB for fixing this!

PR #455, opened 2022-02-23
Removed obsolete bitfields

Silabs-ArjanB:
SEC clean


silabs-robin (approved):
Approval rationale:
* The struct field "uie" is removed as expected.
* "tw" field is added, presumably for 40s merging-compatibility, defaulting to 0 as per user manual.
* ~~I don't know why the parameters are removed, but I'll take that as an implementation detail.~~ I also agree with removing the parameters for the irrelevant S-mode and non-existing U-mode fields.

Issue #452, opened 2022-02-22
RVFI: pc_wdata does not take in to account if a branch is taken when setting a branch address


silabs-mateilga:
cv32e40x/bhv/cv32e40x_rvfi.sv
the "EX STAGE" structure sets pc_wdata, but does not take in to account if a branch is taken or not.

@silabs-halfdan

silabs-halfdan:
This is now fixed.

PR #456, opened 2022-02-24
RVFI bugfixes

silabs-halfdan:
Fixed rvfi issue where rvfi_intr could get set due to an mret that is killed before getting to writeback. 
Fixed rvfi branch handing (#452 )

Also fixed formal warning in load-store unit.


Silabs-ArjanB (approved):
It would have been cleaner to fix the pc_mux_exception related issue in the RTL instead. I will file a separate issue for that and merge this first as it is blocking progress otherwise.

PR #457, opened 2022-02-24
Fixed dsim/vsim compatibility

silabs-halfdan:
Updated sim tracer to use explicit casting to avoid causing issues for some simulators.

### DIFF
--- a/bhv/rvfi_sim_trace.sv
+++ b/bhv/rvfi_sim_trace.sv
@@ -57,7 +57,7 @@ module rvfi_sim_trace
   string              filename, line;
   string              asms[5];
   bit                 itb_file_ok;
-  
+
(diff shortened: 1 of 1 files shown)


silabs-halfdan:
This is a fix for https://github.com/openhwgroup/cv32e40s/issues/148

MikeOpenHWGroup:
Works with dsim.

MikeOpenHWGroup:
Ah, we've got  the same issue on _both_ the E40X and E40S.  This fix also needs to be applied to the E40S.

Silabs-ArjanB:
@MikeOpenHWGroup Yes, that was understood. The fix from the 40X will get merged into the 40S once we do the next merge cycle.

PR #465, opened 2022-02-28
Syntax fix + IF stage fix

Silabs-ArjanB:


Silabs-ArjanB:
Now SEC clean with original if X_EXT = 0

silabs-oysteink (approved):

PR #468, opened 2022-03-03
NMI address and update

silabs-halfdan:
Conditionally added features (pending ISS updates):
 11 bit exception_cause
 fixed offset from mtvec for nmi address 

### DIFF
--- a/rtl/cv32e40x_cs_registers.sv
+++ b/rtl/cv32e40x_cs_registers.sv
@@ -50,7 +50,7 @@ module cv32e40x_cs_registers import cv32e40x_pkg::*;
   input  logic  [3:0]     mimpid_patch_i,
 
   // MTVEC
-  output logic [23:0]     mtvec_addr_o,
+  output logic [21:0]     mtvec_addr_o,
   output logic  [1:0]     mtvec_mode_o,
   
   // Cycle Count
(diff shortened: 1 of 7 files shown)


Silabs-ArjanB (changes requested):
Rename USE_DEPRECAED_FEATURE_SET to USE_DEPRECATED_FEATURE_SET.
Where will this define get set? 
mtvec updates are not according to the user manual.

silabs-halfdan:
Restructured parameterization and implemented mtvec correctly.  Fixed the parameter name typo.
The parameter is now passed to core top-level.

rtl/cv32e40x_cs_registers.sv
```diff
@@ -532,12 +533,14 @@ module cv32e40x_cs_registers import cv32e40x_pkg::*;
...
-                                  exception_code: csr_wdata_int[7:0],
+                                  exception_code: csr_wdata_int[10:0],
                                   default:        'b0
                                 };
     mcause_we                = 1'b0;
-
-    mtvec_n.addr             = csr_mtvec_init_i ? mtvec_addr_i[31:8] : csr_wdata_int[31:8];
+    mtvec_n.addr             = csr_mtvec_init_i ? mtvec_addr_i[31:7] : csr_wdata_int[31:7];
+    if (USE_DEPRECATED_FEATURE_SET) begin
+      mtvec_n[7]             = 1'b0; // todo : remove
```
Silabs-ArjanB, line 542: mtvec_n.addr[7] ?

Silabs-ArjanB (changes requested):

rtl/cv32e40x_cs_registers.sv
silabs-halfdan: Yes, that would give the same result. I just didn't like the non-zero indexing for an lsb of a field.
But I can change it if you think it is more clear.

Silabs-ArjanB: Yes, I think that would be more clear as from above code in isolation I cannot see that this is about the addr part

silabs-halfdan: Ok, pushed an update now.

PR #484, opened 2022-03-23
CLIC: Spec 5.1

silabs-oysteink:
mstatus.mpp and mstatus.mpie is now accessible (RW) via mcause as well as mstatus.

SEC clean when SMCLIC=0


silabs-oysteink:
Spotted mistake, closing PR.

PR #485, opened 2022-03-23
CLIC: Spec chapter 5.1

silabs-oysteink:
Making mstatus.mpie and mstatus.mpp accessible via mcause.

SEC clean with SMCLIC=0

### DIFF
--- a/rtl/cv32e40x_cs_registers.sv
+++ b/rtl/cv32e40x_cs_registers.sv
@@ -534,6 +534,8 @@ module cv32e40x_cs_registers import cv32e40x_pkg::*;
     dscratch1_n              = csr_wdata_int;
     dscratch1_we             = 1'b0;
 
+    // todo: when SD/XS/FS/VS are implemented, a write to mcause.mpp/mpie
+    //       must be handled correctly to not corrupt the mentioned bitfields.
     mstatus_n                = '{
                               tw:   1'b0,
                               mprv: 1'b0,
--- a/rtl/cv32e40x_cs_registers.sv
+++ b/rtl/cv32e40x_cs_registers.sv
@@ -534,16 +534,33 @@ module cv32e40x_cs_registers import cv32e40x_pkg::*;
     dscratch1_n              = csr_wdata_int;
     dscratch1_we             = 1'b0;
 
-    // todo: when SD/XS/FS/VS are implemented, a write to mcause.mpp/mpie
-    //       must be handled correctly to not corrupt the mentioned bitfields.
-    mstatus_n                = '{
-                              tw:   1'b0,
-                              mprv: 1'b0,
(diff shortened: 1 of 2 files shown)


rtl/cv32e40x_cs_registers.sv
```diff
@@ -306,7 +306,20 @@ module cv32e40x_cs_registers import cv32e40x_pkg::*;
       CSR_MEPC: csr_rdata_int = mepc_q;
 
       // mcause: exception cause
-      CSR_MCAUSE: csr_rdata_int = mcause_q;
+      CSR_MCAUSE: begin
+        if (SMCLIC) begin
+          // For CLIC, the mpp and mpie bits from mstatus
+          // are readable via mcause
```
Silabs-ArjanB, line 312: Add note that 'CLIC mode is assumed when SMCLIC = 1'. RTL should actually have checked on 'clic mode' instead of SMCLIC, but let's keep it like this as these two things are the same for our subset of the Smclic implementation (and your implementation is better for coverage reasons)

rtl/cv32e40x_cs_registers.sv
```diff
@@ -521,16 +534,36 @@ module cv32e40x_cs_registers import cv32e40x_pkg::*;
...
-    mstatus_n                = '{
-                              tw:   1'b0,
-                              mprv: 1'b0,
-                              mpp:  PRIV_LVL_M,
-                              mpie: csr_wdata_int[MSTATUS_MPIE_BIT],
-                              mie:  csr_wdata_int[MSTATUS_MIE_BIT],
-                              default: 'b0
-                            };
+    if (SMCLIC) begin
+      // SMCLIC: A write to mcause.mpp or mcause.mpie is stored in mstatus
```
Silabs-ArjanB, line 538: Same remark as above: 'CLIC mode is assumed when SMCLIC = 1'.

rtl/cv32e40x_cs_registers.sv
```diff
@@ -521,16 +534,36 @@ module cv32e40x_cs_registers import cv32e40x_pkg::*;
...
+      if(mcause_we) begin
+        mstatus_n           = mstatus_q;
+        mstatus_n.mpp       = PRIV_LVL_M;
+        mstatus_n.mpie      = csr_wdata_int[MCAUSE_MPIE_BIT];
+      end else begin
+        // TODO: add support for SD/XS/FS/VS
+        mstatus_n              = '{
```
Silabs-ArjanB, line 545: Reorganize code to avoid this code duplication. Simply first do the original mstatus_n assignment and then a if (SMCLIC) if (mcause_we)

Silabs-ArjanB (changes requested):

Silabs-ArjanB (approved):

PR #486, opened 2022-03-24
CLIC: Spec chapter 5.3

silabs-oysteink:
CSRs mie and mip appear as hardwired to zero when SMCLIC = 1.
Writes are ignored.

SEC clean when SMCLIC = 0


Silabs-ArjanB (changes requested):
All the code is good, but I would go a bit further and do the following as well:

Use generate statement around MIE csr.
Use generate statement around most/all? code in int_controller (likely it will al get optimized away, but we want it gone for code coverage reasons as well).
Add assertions at core level that mip and mie are 0 when SMCLIC=1

Silabs-ArjanB (approved):

PR #487, opened 2022-03-24
RVFI Sleep Signals

silabs-halfdan:
Added rvfi_wu and rvfi_sleep signals to rvfi, fixed issue with rvfi_csr_mip_rdata signal

### DIFF
--- a/bhv/cv32e40x_rvfi.sv
+++ b/bhv/cv32e40x_rvfi.sv
@@ -75,6 +75,7 @@ module cv32e40x_rvfi
    input logic                                ebreak_in_wb_i,
    input logic [31:0]                         instr_rdata_wb_i,
    input logic                                csr_en_wb_i,
+   input logic                                sys_wfi_insn_wb_i,
    // Register writes
    input logic                                rf_we_wb_i,
    input logic [4:0]                          rf_addr_wb_i,
@@ -98,6 +99,7 @@ module cv32e40x_rvfi
    input logic                                debug_mode_q_i,
 
    // Interrupt Controller probes
+   input logic [31:0]                         irq_i,
    input logic                                irq_wu_ctrl_i,
    input logic [4:0]                          irq_id_ctrl_i,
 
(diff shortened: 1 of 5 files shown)


bhv/cv32e40x_rvfi.sv
```diff
@@ -965,7 +1010,7 @@ module cv32e40x_rvfi
   assign rvfi_csr_wmask_d.mtval              = '0;
 
   assign ex_csr_rdata_d.mip                  = csr_mip_q_i;
-  assign rvfi_csr_rdata_d.mip                = ex_csr_rdata.mip;
+  assign rvfi_csr_rdata_d.mip                = csr_en_wb_i ? ex_csr_rdata.mip : csr_mip_q_i;
```
Silabs-ArjanB, line 1013: Please add a comment on 'why' this was done

bhv/cv32e40x_rvfi.sv
```diff
@@ -817,11 +857,16 @@ module cv32e40x_rvfi
         rvfi_dbg       <= debug_cause[STAGE_WB];
         rvfi_dbg_mode  <= debug_mode [STAGE_WB];
 
+        rvfi_sleep     <= (ctrl_fsm_ns_i == SLEEP);
```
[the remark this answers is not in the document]
silabs-halfdan: Good point, WFIs will not enter sleep if there are flopped wakeup causes. This code will signal a sleep and wakeup event if the wakeup cause is set in the same cycle as the WFI retires, although the core_sleep_o signal will never be set.
I can update it to also consider the un-flopped wakeup signals.

bhv/cv32e40x_rvfi.sv
silabs-halfdan: Will add a comment

Silabs-ArjanB, line 1013: Is this also correct for WFI instructions, which take (at least) 2 cycles in WB and for which I believe they are reported on RVFI during their first WB cycle (so not necessarily with the correct MIP I think)

Silabs-ArjanB (changes requested):

Silabs-ArjanB (approved):

PR #492, opened 2022-03-29
Basic CLIC vectoring support

silabs-oysteink:
Work-in-progress, do not merge.

Basic CLIC controller (without thresholds) is included.
- For CLIC, vectoring is implemented through the alignment buffer.
Core should still work in basic mode.

### DIFF
--- a/rtl/cv32e40x_alignment_buffer.sv
+++ b/rtl/cv32e40x_alignment_buffer.sv
@@ -116,6 +116,9 @@ module cv32e40x_alignment_buffer import cv32e40x_pkg::*;
   // Request a transfer when needed, or we do a branch, iff outstanding_cnt_q is less than 2
   // Adjust instruction count with 'pop_q', which tells if we consumed an instruction
   // during the last cycle
+  // For CLIC vector loads, if the initial pc_set does not immediately get a fetch_ready,
+  // ptr_fetch_done_q will remain low to enable the prefetcher to fetch the pointer even though
+  // the controller pulls instr_req low while waiting for the pointer result.
   assign fetch_valid_o = (ctrl_fsm_i.instr_req || !ptr_fetch_done_q) &&
                          (outstanding_cnt_q < 2) &&
                          (((instr_cnt_q - pop_q) == 'd0) ||
@@ -566,5 +569,5 @@ module cv32e40x_alignment_buffer import cv32e40x_pkg::*;
   // If the pointer fetch isn't granted immediately, is_ptr_q will be true until
   // the cycle after the next pc_set. Must avoid signaling fetch_data_access_o
   // for the first pc_set after pointer fetch.
-  assign fetch_data_access_o = ctrl_fsm_i.pc_set_clicv || (is_ptr_q && !(ctrl_fsm_i.pc_set && !ctrl_fsm_i.pc_set_clicv));
+  assign fetch_data_access_o = (ctrl_fsm_i.pc_set && ctrl_fsm_i.pc_set_clicv) || (is_ptr_q && !(ctrl_fsm_i.pc_set && !ctrl_fsm_i.pc_set_clicv));
 endmodule
--- a/rtl/cv32e40x_alignment_buffer.sv
+++ b/rtl/cv32e40x_alignment_buffer.sv
@@ -39,6 +39,7 @@ module cv32e40x_alignment_buffer import cv32e40x_pkg::*;
   input  logic           fetch_ready_i,
   output logic           fetch_branch_o,
   output logic [31:0]    fetch_branch_addr_o,
+  output logic           fetch_data_access_o,
 
   // Resp interface
   input  logic           resp_valid_i,
@@ -49,7 +50,8 @@ module cv32e40x_alignment_buffer import cv32e40x_pkg::*;
   output logic           instr_valid_o,
   input  logic           instr_ready_i,
   output inst_resp_t     instr_instr_o,
-  output logic [31:0]    instr_addr_o
+  output logic [31:0]    instr_addr_o,
+  output logic           instr_is_ptr_o
 
 );
 
@@ -99,6 +101,11 @@ module cv32e40x_alignment_buffer import cv32e40x_pkg::*;
   // resp_valid gated while flushing
   logic resp_valid_gated;
 
+  // CLIC vectoring (and Zce table jumps)
+  // Flag for signalling that results is a function pointer
+  logic is_ptr_q;
+  logic ptr_fetch_done_q;
+
   assign resp_valid_gated = (n_flush_q > 0) ? 1'b0 : resp_valid_i;
 
   // Assume we always push all instructions to FIFO
@@ -109,7 +116,7 @@ module cv32e40x_alignment_buffer import cv32e40x_pkg::*;
   // Request a transfer when needed, or we do a branch, iff outstanding_cnt_q is less than 2
   // Adjust instruction count with 'pop_q', which tells if we consumed an instruction
   // during the last cycle
-  assign fetch_valid_o = ctrl_fsm_i.instr_req &&
+  assign fetch_valid_o = (ctrl_fsm_i.instr_req || !ptr_fetch_done_q) &&
                          (outstanding_cnt_q < 2) &&
                          (((instr_cnt_q - pop_q) == 'd0) ||
                          ((instr_cnt_q - pop_q) == 'd1 && outstanding_cnt_q == 2'd0) ||
@@ -472,17 +479,19 @@ module cv32e40x_alignment_buffer import cv32e40x_pkg::*;
   begin
     if(rst_n == 1'b0)
     begin
-      addr_q    <= '0;
-      resp_q    <= INST_RESP_RESET_VAL;
-      valid_q   <= '0;
-      aligned_q <= 1'b0;
-      complete_q <= 1'b0;
-      n_flush_q  <= 'd0;
-      instr_cnt_q <= 'd0;
+      addr_q            <= '0;
+      resp_q            <= INST_RESP_RESET_VAL;
+      valid_q           <= '0;
+      aligned_q         <= 1'b0;
+      complete_q        <= 1'b0;
+      n_flush_q         <= 'd0;
+      instr_cnt_q       <= 'd0;
       outstanding_cnt_q <= 'd0;
-      rptr      <= 'd0;
-      wptr      <= 'd0;
-      pop_q     <= 1'b0;
+      rptr              <= 'd0;
+      wptr              <= 'd0;
+      pop_q             <= 1'b0;
+      is_ptr_q          <= 1'b0;
+      ptr_fetch_done_q  <= 1'b1; // Reset to 1 to avoid pointer fetches out of reset
     end
     else
     begin
@@ -508,6 +517,7 @@ module cv32e40x_alignment_buffer import cv32e40x_pkg::*;
         // Reset pointers on branch
         wptr <= 'd0;
         rptr <= 'd0;
+        is_ptr_q <= ctrl_fsm_i.pc_set_clicv;
       end else begin
         // Update write pointer on a valid response
         if (resp_valid_gated) begin
@@ -521,6 +531,18 @@ module cv32e40x_alignment_buffer import cv32e40x_pkg::*;
         end
       end
 
+      // Set pointer fetch done flag when a fetch is granted
+      // If the fetch is accepted the same cycle as the pointer request,
+      // the flag will remain high.
+      if(fetch_valid_o && fetch_ready_i) begin
+        ptr_fetch_done_q <= 1'b1;
+      end else begin
+        // Clear flag if no fetch is granted this cycle, but we do have a pointer request
+        // This ensures that we keep setting fetch_valid_o for the pointer until it is accepted.
+        if(ctrl_fsm_i.pc_set && ctrl_fsm_i.pc_set_clicv) begin
+          ptr_fetch_done_q <= 1'b0;
+        end
+      end
       // Set pop-bit when instruction is emitted.
       pop_q <= (instr_valid_o && instr_ready_i);
 
@@ -535,4 +557,14 @@ module cv32e40x_alignment_buffer import cv32e40x_pkg::*;
   // Output instruction address to if_stage
   assign instr_addr_o = addr_q;
 
+  // Signal that result is a pointer
+  // CLIC vectoring or Zce table jump
+  assign instr_is_ptr_o = is_ptr_q;
+
+  // Fetch is treated as a data access for CLIC vector fetch
+  // Only set for the initial pc_set_clicv point fetch
+  // If the pointer fetch isn't granted immediately, is_ptr_q will be true until
+  // the cycle after the next pc_set. Must avoid signaling fetch_data_access_o
+  // for the first pc_set after pointer fetch.
+  assign fetch_data_access_o = ctrl_fsm_i.pc_set_clicv || (is_ptr_q && !(ctrl_fsm_i.pc_set && !ctrl_fsm_i.pc_set_clicv));
 endmodule
--- a/rtl/cv32e40x_alignment_buffer.sv
+++ b/rtl/cv32e40x_alignment_buffer.sv
@@ -40,6 +40,7 @@ module cv32e40x_alignment_buffer import cv32e40x_pkg::*;
   output logic           fetch_branch_o,
   output logic [31:0]    fetch_branch_addr_o,
   output logic           fetch_data_access_o,
+  input  logic           fetch_ptr_access_i,
 
   // Resp interface
   input  logic           resp_valid_i,
@@ -104,7 +105,7 @@ module cv32e40x_alignment_buffer import cv32e40x_pkg::*;
   // CLIC vectoring (and Zc table jumps)
   // Flag for signalling that results is a function pointer
   logic is_ptr_q;
-  logic ptr_fetch_done_q;
+  logic ptr_fetch_accepted_q;
 
   assign resp_valid_gated = (n_flush_q > 0) ? 1'b0 : resp_valid_i;
 
@@ -116,12 +117,12 @@ module cv32e40x_alignment_buffer import cv32e40x_pkg::*;
   // Request a transfer when needed, or we do a branch, iff outstanding_cnt_q is less than 2
   // Adjust instruction count with 'pop_q', which tells if we consumed an instruction
   // during the last cycle
-  // For CLIC vector loads, if the initial pc_set does not immediately get a fetch_ready,
-  // ptr_fetch_done_q will remain low to enable the prefetcher to fetch the pointer even though
-  // the controller pulls instr_req low while waiting for the pointer result.
-  assign fetch_valid_o = (ctrl_fsm_i.instr_req || !ptr_fetch_done_q) &&
-                         (outstanding_cnt_q < 2) &&
-                         (((instr_cnt_q - pop_q) == 'd0) ||
+  // For CLIC vector loads we want to stop prefetching between an accepted pointer load and
+  // the next pc_set (to the target address).
+  assign fetch_valid_o = (ctrl_fsm_i.instr_req )                                     &&
+                         (outstanding_cnt_q < 2)                                     &&
+                         !(ptr_fetch_accepted_q && !ctrl_fsm_i.pc_set)               && // No fetch until next pc_set after accepted pointer fetches
+                         (((instr_cnt_q - pop_q) == 'd0)                             ||
                          ((instr_cnt_q - pop_q) == 'd1 && outstanding_cnt_q == 2'd0) ||
                          ctrl_fsm_i.pc_set);
 
@@ -494,7 +495,7 @@ module cv32e40x_alignment_buffer import cv32e40x_pkg::*;
       wptr              <= 'd0;
       pop_q             <= 1'b0;
       is_ptr_q          <= 1'b0;
-      ptr_fetch_done_q  <= 1'b1; // Reset to 1 to avoid pointer fetches out of reset
+      ptr_fetch_accepted_q  <= 1'b0;
     end
     else
     begin
@@ -515,6 +516,9 @@ module cv32e40x_alignment_buffer import cv32e40x_pkg::*;
       end
 
       // Update address and read/write pointers on a requested branch
+      // Incoming pointers for CLIC and Zc may cause the pointers to get wrong values
+      // We do however prevent further fetches after a pointer fetch, and the pointers
+      // will be reset on the next pc_set (jump to target)
       if (ctrl_fsm_i.pc_set) begin
         addr_q  <= branch_addr_i;       // Branch target address will correspond to first instruction received after this.
         // Reset pointers on branch
@@ -534,16 +538,11 @@ module cv32e40x_alignment_buffer import cv32e40x_pkg::*;
         end
       end
 
-      // Set pointer fetch done flag when a fetch is granted
-      // If the fetch is accepted the same cycle as the pointer request,
-      // the flag will remain high.
-      if(fetch_valid_o && fetch_ready_i) begin
-        ptr_fetch_done_q <= 1'b1;
+      if(fetch_valid_o && fetch_ready_i && fetch_ptr_access_i) begin
+        ptr_fetch_accepted_q <= 1'b1;
       end else begin
-        // Clear flag if no fetch is granted this cycle, but we do have a pointer request
-        // This ensures that we keep setting fetch_valid_o for the pointer until it is accepted.
-        if(ctrl_fsm_i.pc_set && ctrl_fsm_i.pc_set_clicv) begin
-          ptr_fetch_done_q <= 1'b0;
+        if(ctrl_fsm_i.pc_set) begin
+          ptr_fetch_accepted_q <= 1'b0;
         end
       end
       // Set pop-bit when instruction is emitted.
@@ -565,9 +564,7 @@ module cv32e40x_alignment_buffer import cv32e40x_pkg::*;
   assign instr_is_ptr_o = is_ptr_q;
 
   // Fetch is treated as a data access for CLIC vector fetch
-  // Only set for the initial pc_set_clicv point fetch
-  // If the pointer fetch isn't granted immediately, is_ptr_q will be true until
-  // the cycle after the next pc_set. Must avoid signaling fetch_data_access_o
-  // for the first pc_set after pointer fetch.
-  assign fetch_data_access_o = (ctrl_fsm_i.pc_set && ctrl_fsm_i.pc_set_clicv) || (is_ptr_q && !(ctrl_fsm_i.pc_set && !ctrl_fsm_i.pc_set_clicv));
+  // Only set for the initial pc_set_clicv point fetch and qualified with fetch_branch_o
+  // Prefetcher will remember the state of data_access in case of stalls.
+  assign fetch_data_access_o = (ctrl_fsm_i.pc_set && ctrl_fsm_i.pc_set_clicv);
 endmodule
--- a/rtl/cv32e40x_alignment_buffer.sv
+++ b/rtl/cv32e40x_alignment_buffer.sv
@@ -101,7 +101,7 @@ module cv32e40x_alignment_buffer import cv32e40x_pkg::*;
   // resp_valid gated while flushing
   logic resp_valid_gated;
 
-  // CLIC vectoring (and Zce table jumps)
+  // CLIC vectoring (and Zc table jumps)
   // Flag for signalling that results is a function pointer
   logic is_ptr_q;
   logic ptr_fetch_done_q;
@@ -558,7 +558,7 @@ module cv32e40x_alignment_buffer import cv32e40x_pkg::*;
   assign instr_addr_o = addr_q;
 
   // Signal that result is a pointer
-  // CLIC vectoring or Zce table jump
+  // CLIC vectoring or Zc table jump
   assign instr_is_ptr_o = is_ptr_q;
 
   // Fetch is treated as a data access for CLIC vector fetch
--- a/rtl/cv32e40x_clic_int_controller.sv
+++ b/rtl/cv32e40x_clic_int_controller.sv
@@ -1,0 +1,111 @@
+// Copyright 2022 Silicon Labs, Inc.
+//
+// This file, and derivatives thereof are licensed under the
+// Solderpad License, Version 2.0 (the "License");
+// Use of this file means you agree to the terms and conditions
+// of the license and are in full compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://solderpad.org/licenses/SHL-2.0/
+//
+// Unless required by applicable law or agreed to in writing, software
+// and hardware implementations thereof
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, EITHER EXPRESSED OR IMPLIED.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+////////////////////////////////////////////////////////////////////////////////
+// Engineers       Oystein Knauserud -     oystein.knauserud@silabs.com       //
+//                                                                            //
+// Design Name:    CLIC int controller                                        //
+// Project Name:   CV32E40X                                                   //
+// Language:       SystemVerilog                                              //
+//                                                                            //
+// Description:    Controller for handling CLIC interrupts                    //
+//                                                                            //
+////////////////////////////////////////////////////////////////////////////////
+
+module cv32e40x_clic_int_controller import cv32e40x_pkg::*;
+#(
+    parameter int SMCLIC_ID_WIDTH = 5
+)
+(
+  input  logic                       clk,
+  input  logic                       rst_n,
+
+  // CLIC interface
+  input  logic                       clic_irq_i,       // CLIC interrupt pending
+  input  logic [SMCLIC_ID_WIDTH-1:0] clic_irq_id_i,    // ID of pending interrupt
+  input  logic [7:0]                 clic_irq_level_i, // Level of pending interrupt
+  input  logic [1:0]                 clic_irq_priv_i,  // Privilege level of pending interrupt
+  input  logic                       clic_irq_shv_i,   // Is pending interrupt vectored?
+
+
+  // To cv32e40x_controller
+  output logic                       irq_req_ctrl_o,
+  output logic [SMCLIC_ID_WIDTH-1:0] irq_id_ctrl_o,
+  output logic                       irq_wu_ctrl_o,
+
+  // To/from cv32e40x_cs_registers
+  input  logic                       m_ie_i             // Interrupt enable bit from CSR (M mode)
+);
+
+  logic                       global_irq_enable;
+
+  // Flops for breaking timing path to instruction interface
+  logic                       clic_irq_q;
+  logic [SMCLIC_ID_WIDTH-1:0] clic_irq_id_q;
+  logic [7:0]                 clic_irq_level_q;
+  logic [1:0]                 clic_irq_priv_q;
+  logic                       clic_irq_shv_q;
+
+  // Register interrupt input (on gated clock). The wake-up logic will
+  // observe clic_irq_i as well, but in all other places clic_irq_q will be used to
+  // avoid timing paths from clic_irq_i to instr_*_o
+
+  always_ff @(posedge clk, negedge rst_n)
+  begin
+    if (rst_n == 1'b0) begin
+      clic_irq_q     <= 1'b0;
+    end else begin
+      clic_irq_q     <= clic_irq_i;
+    end
+  end
+
+
+  always_ff @(posedge clk, negedge rst_n)
+  begin
+    if (rst_n == 1'b0) begin
+      clic_irq_id_q     <= '0;
+      clic_irq_level_q  <= '0;
+      clic_irq_priv_q   <= '0;
+      clic_irq_shv_q    <= 1'b0;
+    end else begin
+      if (clic_irq_i) begin
+        clic_irq_id_q    <= clic_irq_id_i;
+        clic_irq_level_q <= clic_irq_level_i;
+        clic_irq_priv_q  <= clic_irq_priv_i;
+        clic_irq_shv_q   <= clic_irq_shv_i;
+      end
+    end
+  end
+
+    // Global interrupt enable
+  assign global_irq_enable = m_ie_i;
+
+  ///////////////////////////
+  // Outputs to controller //
+  ///////////////////////////
+
+  // Request to take interrupt if there a pending-and-enabled interrupt and interrupts are enabled globally
+  // todo: factor in interrupt level and thresholds
+  assign irq_req_ctrl_o = clic_irq_q && global_irq_enable;
+
+  // Pass on interrupt ID
+  assign irq_id_ctrl_o = clic_irq_id_q;
+
+  // Wake-up signal based on unregistered IRQ such that wake-up can be caused if no clock is present
+  assign irq_wu_ctrl_o = clic_irq_i;
+
+endmodule // cv32e40x_clic_int_controller
--- a/rtl/cv32e40x_clic_int_controller.sv
+++ b/rtl/cv32e40x_clic_int_controller.sv
@@ -44,7 +44,7 @@ module cv32e40x_clic_int_controller import cv32e40x_pkg::*;
 
   // To cv32e40x_controller
   output logic                       irq_req_ctrl_o,
-  output logic [SMCLIC_ID_WIDTH-1:0] irq_id_ctrl_o,
+  output logic [9:0]                 irq_id_ctrl_o,    // Max width - unused bits are tied off
   output logic                       irq_wu_ctrl_o,
   output logic                       irq_clic_shv_o,
   output logic [7:0]                 irq_clic_level_o,
@@ -61,7 +61,7 @@ module cv32e40x_clic_int_controller import cv32e40x_pkg::*;
 
   // Flops for breaking timing path to instruction interface
   logic                       clic_irq_q;
-  logic [SMCLIC_ID_WIDTH-1:0] clic_irq_id_q;
+  logic [9:0]                 clic_irq_id_q;
   logic [7:0]                 clic_irq_level_q;
   logic [1:0]                 clic_irq_priv_q;
   logic                       clic_irq_shv_q;
@@ -89,8 +89,8 @@ module cv32e40x_clic_int_controller import cv32e40x_pkg::*;
       clic_irq_shv_q    <= 1'b0;
     end else begin
       if (clic_irq_i) begin
-        clic_irq_id_q    <= clic_irq_id_i;
-        clic_irq_level_q <= clic_irq_level_i;
+        clic_irq_id_q    <= 10'(clic_irq_id_i); // Casting SMCLIC_ID_WIDTH into max with of 10 bits.
+        clic_irq_level_q <= clic_irq_level_i;   // Will always be PRIV_LVL_M todo: add assertion
         clic_irq_priv_q  <= clic_irq_priv_i;
         clic_irq_shv_q   <= clic_irq_shv_i;
       end
--- a/rtl/cv32e40x_clic_int_controller.sv
+++ b/rtl/cv32e40x_clic_int_controller.sv
@@ -47,12 +47,17 @@ module cv32e40x_clic_int_controller import cv32e40x_pkg::*;
   output logic [SMCLIC_ID_WIDTH-1:0] irq_id_ctrl_o,
   output logic                       irq_wu_ctrl_o,
   output logic                       irq_clic_shv_o,
+  output logic [7:0]                 irq_clic_level_o,
 
-  // To/from cv32e40x_cs_registers
-  input  logic                       m_ie_i             // Interrupt enable bit from CSR (M mode)
+  // From cv32e40x_cs_registers
+  input  logic                       m_ie_i,             // Interrupt enable bit from CSR (M mode)
+  input  logic [7:0]                 mintthresh_i,       // Current interrupt threshold from CSR
+  input  mintstatus_t                mintstatus_i        // Current mintstatus from CSR
 );
 
   logic                       global_irq_enable;
+  logic  [7:0]                effective_irq_level; // Calculate effective interrupt level
+
 
   // Flops for breaking timing path to instruction interface
   logic                       clic_irq_q;
@@ -92,23 +97,40 @@ module cv32e40x_clic_int_controller import cv32e40x_pkg::*;
     end
   end
 
-    // Global interrupt enable
+  // Global interrupt enable
+  // todo: move logic for m_ie_i from cs_registers to here.
   assign global_irq_enable = m_ie_i;
 
+  assign effective_irq_level = (mintthresh_i > mintstatus_i.mil) ? mintthresh_i : mintstatus_i.mil;
   ///////////////////////////
   // Outputs to controller //
   ///////////////////////////
 
-  // Request to take interrupt if there a pending-and-enabled interrupt and interrupts are enabled globally
-  // todo: factor in interrupt level and thresholds
-  assign irq_req_ctrl_o = clic_irq_q && global_irq_enable;
+  // Request to take interrupt if:
+  // There a pending-and-enabled interrupt and interrupts are enabled globally
+  // AND the incoming irq level is above the cores current effective interrupt level.
+  // todo: In user mode, machine threshold should not mask interrupts to machine mode
+  assign irq_req_ctrl_o = clic_irq_q &&
+                          clic_irq_level_q > effective_irq_level &&
+                          global_irq_enable;
 
   // Pass on interrupt ID
   assign irq_id_ctrl_o = clic_irq_id_q;
 
   // Wake-up signal based on unregistered IRQ such that wake-up can be caused if no clock is present
-  assign irq_wu_ctrl_o = clic_irq_i;
+  // SMCLIC spec states three scenarios for wakeup:
+  // 1: priv mode  > current, irq i is max (done in external CLIC), level != 0
+  // 2: priv mode == current, irq i is max (done in external CLIC), level > max(mintstatus.mil, mintthresh.th)
+  // 3: priv mode  < current, irq_i is max (done in external CLIC), level != 0
+  //
+  // 1 is applicable for E40S only as E40X only runs in machine mode
+  // 2 is applicable for both E40S and E40X
+  // 3 is not applicable, we support machine mode interrupts only.
+  // todo: implement (2) for E40S.
+  assign irq_wu_ctrl_o = clic_irq_i && (clic_irq_level_i > effective_irq_level);
 
   assign irq_clic_shv_o = clic_irq_shv_q;
+
+  assign irq_clic_level_o = clic_irq_level_q;
 
 endmodule // cv32e40x_clic_int_controller
--- a/rtl/cv32e40x_clic_int_controller.sv
+++ b/rtl/cv32e40x_clic_int_controller.sv
@@ -46,6 +46,7 @@ module cv32e40x_clic_int_controller import cv32e40x_pkg::*;
   output logic                       irq_req_ctrl_o,
   output logic [SMCLIC_ID_WIDTH-1:0] irq_id_ctrl_o,
   output logic                       irq_wu_ctrl_o,
+  output logic                       irq_clic_shv_o,
 
   // To/from cv32e40x_cs_registers
   input  logic                       m_ie_i             // Interrupt enable bit from CSR (M mode)
@@ -67,9 +68,9 @@ module cv32e40x_clic_int_controller import cv32e40x_pkg::*;
   always_ff @(posedge clk, negedge rst_n)
   begin
     if (rst_n == 1'b0) begin
-      clic_irq_q     <= 1'b0;
+      clic_irq_q  <= 1'b0;
     end else begin
-      clic_irq_q     <= clic_irq_i;
+      clic_irq_q  <= clic_irq_i;
     end
   end
 
@@ -79,7 +80,7 @@ module cv32e40x_clic_int_controller import cv32e40x_pkg::*;
     if (rst_n == 1'b0) begin
       clic_irq_id_q     <= '0;
       clic_irq_level_q  <= '0;
-      clic_irq_priv_q   <= '0;
+      clic_irq_priv_q   <= PRIV_LVL_M;
       clic_irq_shv_q    <= 1'b0;
     end else begin
       if (clic_irq_i) begin
@@ -107,5 +108,7 @@ module cv32e40x_clic_int_controller import cv32e40x_pkg::*;
 
   // Wake-up signal based on unregistered IRQ such that wake-up can be caused if no clock is present
   assign irq_wu_ctrl_o = clic_irq_i;
+
+  assign irq_clic_shv_o = clic_irq_shv_q;
 
 endmodule // cv32e40x_clic_int_controller
--- a/rtl/cv32e40x_clic_int_controller.sv
+++ b/rtl/cv32e40x_clic_int_controller.sv
@@ -107,11 +107,11 @@ module cv32e40x_clic_int_controller import cv32e40x_pkg::*;
   ///////////////////////////
 
   // Request to take interrupt if:
-  // There a pending-and-enabled interrupt and interrupts are enabled globally
-  // AND the incoming irq level is above the cores current effective interrupt level.
+  // There is a pending-and-enabled interrupt and interrupts are enabled globally
+  // AND the incoming irq level is above the core's current effective interrupt level.
   // todo: In user mode, machine threshold should not mask interrupts to machine mode
   assign irq_req_ctrl_o = clic_irq_q &&
-                          clic_irq_level_q > effective_irq_level &&
+                          (clic_irq_level_q > effective_irq_level) &&
                           global_irq_enable;
 
   // Pass on interrupt ID
--- a/rtl/cv32e40x_clic_int_controller.sv
+++ b/rtl/cv32e40x_clic_int_controller.sv
@@ -127,6 +127,7 @@ module cv32e40x_clic_int_controller import cv32e40x_pkg::*;
   // 2 is applicable for both E40S and E40X
   // 3 is not applicable, we support machine mode interrupts only.
   // todo: implement (2) for E40S.
+  // todo: can we share the comparator below and flop the result for irq_req_ctrl_o?
   assign irq_wu_ctrl_o = clic_irq_i && (clic_irq_level_i > effective_irq_level);
 
   assign irq_clic_shv_o = clic_irq_shv_q;
--- a/rtl/cv32e40x_controller_fsm.sv
+++ b/rtl/cv32e40x_controller_fsm.sv
@@ -56,7 +56,7 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
   input  logic        sys_mret_id_i,              // mret in ID stage
 
   // From EX stage
-  input  id_ex_pipe_t id_ex_pipe_i,        
+  input  id_ex_pipe_t id_ex_pipe_i,
   input  logic        branch_decision_ex_i,       // branch decision signal from EX ALU
   input  logic        lsu_split_ex_i,             // LSU is splitting misaligned, first half is in EX
 
@@ -129,7 +129,7 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
   logic single_step_halt_if_n;
   logic single_step_halt_if_q; // Halting IF after issuing one insn in single step mode
 
-  
+
   // Events in ID
   logic jump_in_id;
   logic jump_taken_id;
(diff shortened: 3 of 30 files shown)


rtl/cv32e40x_alignment_buffer.sv
```diff
@@ -535,4 +557,14 @@ module cv32e40x_alignment_buffer import cv32e40x_pkg::*;
   // Output instruction address to if_stage
   assign instr_addr_o = addr_q;
 
+  // Signal that result is a pointer
+  // CLIC vectoring or Zce table jump
```
Silabs-ArjanB, line 561: Zce -> Zc

rtl/cv32e40x_clic_int_controller.sv
```diff
@@ -0,0 +1,111 @@
...
+  logic                       clic_irq_shv_q;
+
+  // Register interrupt input (on gated clock). The wake-up logic will
+  // observe clic_irq_i as well, but in all other places clic_irq_q will be used to
+  // avoid timing paths from clic_irq_i to instr_*_o
+
+  always_ff @(posedge clk, negedge rst_n)
+  begin
+    if (rst_n == 1'b0) begin
+      clic_irq_q     <= 1'b0;
```
Silabs-ArjanB, line 70: Remove repeated spaces before <=

rtl/cv32e40x_clic_int_controller.sv
```diff
@@ -0,0 +1,111 @@
...
+      clic_irq_id_q     <= '0;
+      clic_irq_level_q  <= '0;
+      clic_irq_priv_q   <= '0;
+      clic_irq_shv_q    <= 1'b0;
+    end else begin
+      if (clic_irq_i) begin
+        clic_irq_id_q    <= clic_irq_id_i;
+        clic_irq_level_q <= clic_irq_level_i;
+        clic_irq_priv_q  <= clic_irq_priv_i;
+        clic_irq_shv_q   <= clic_irq_shv_i;
```
Silabs-ArjanB, line 89: level, priv, shv not used yet.
Actually priv will always be M mode (also for 40S); add assertion to check this and also add comment that this will be the case. Maybe change reset value of priv_q to machine mode so that this will surely get optimized away

rtl/cv32e40x_controller_fsm.sv
```diff
@@ -193,10 +195,10 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
   // Once the fencei handshake is initiated, it must complete and the instruction must retire.
   // The instruction retires when fencei_req_and_ack_q = 1
   assign fencei_ongoing = fencei_flush_req_o || fencei_req_and_ack_q;
-  
+
   // Mux selector for vectored IRQ PC
-  assign ctrl_fsm_o.m_exc_vec_pc_mux = (mtvec_mode_i == 2'b0) ? 5'h0 : exc_cause;
-  
+  assign ctrl_fsm_o.m_exc_vec_pc_mux = (mtvec_mode_i == 2'b0) ? 11'h0 : exc_cause;
```
Silabs-ArjanB, line 200: I think it will be cleaner to have a separate mtvec_pc_mux and mtvt_pc_mux

rtl/cv32e40x_controller_fsm.sv
```diff
@@ -229,6 +231,7 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
                             (lsu_mpu_status_wb_i != MPU_OK)) && ex_wb_pipe_i.instr_valid;
 
   // Set exception cause
+  // todo: CLIC, faulted vector loads (from the IF stage) must be treated as EXC_CAUSE_LOAD_FAULT
```
Silabs-ArjanB, line 234: Can that already be done in the MPU?

rtl/cv32e40x_controller_fsm.sv
```diff
@@ -451,14 +454,15 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
 
     fencei_flush_req_set           = 1'b0;
 
+    ctrl_fsm_o.pc_set_clicv = 1'b0;
```
Silabs-ArjanB, line 457: Not aligned to previous line

rtl/cv32e40x_controller_fsm.sv
```diff
@@ -515,15 +519,25 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
           ctrl_fsm_o.kill_wb = 1'b1;
 
           ctrl_fsm_o.pc_set = 1'b1;
-          ctrl_fsm_o.pc_mux = PC_TRAP_IRQ;
+
           exc_cause = irq_id_ctrl_i;
 
           ctrl_fsm_o.irq_ack = 1'b1;
           ctrl_fsm_o.irq_id  = irq_id_ctrl_i;
```
Silabs-ArjanB, line 526: Width of irq_id, irq_id_ctrl_i should be updated to deal with 1024 interrupts.

rtl/cv32e40x_controller_fsm.sv
```diff
@@ -515,15 +519,25 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
           ctrl_fsm_o.kill_wb = 1'b1;
 
           ctrl_fsm_o.pc_set = 1'b1;
-          ctrl_fsm_o.pc_mux = PC_TRAP_IRQ;
+
           exc_cause = irq_id_ctrl_i;
```
Silabs-ArjanB, line 523: Width mismatch

rtl/cv32e40x_controller_fsm.sv
```diff
@@ -724,6 +738,23 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
         debug_mode_n = 1'b1;
         ctrl_fsm_ns = FUNCTIONAL;
       end
+      // State for CLIC vectoring (and Zce table jumps)
```
Silabs-ArjanB, line 741: Zce -> Zc

rtl/cv32e40x_controller_fsm.sv
```diff
@@ -724,6 +738,23 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
...
+      // In this state a fetch has been ordered, and the controller
+      // is waiting for the pointer to arrive in the decode stage.
+      VECTORING: begin
+        // Stop further prefetching
+        ctrl_fsm_o.instr_req = 1'b0;
+
+        if(if_id_pipe_i.instr_meta.clicv) begin
```
Silabs-ArjanB, line 748: if( -> if (

Silabs-ArjanB, line 748: Shouldn't this be qualified with some valid signal? Also add comment on how associated errors (e.g. bus error) will be handled.

rtl/cv32e40x_controller_fsm.sv
```diff
@@ -724,6 +738,23 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
...
+          // Function pointer reached ID stage, do another jump
+          ctrl_fsm_o.instr_req = 1'b1;
+          ctrl_fsm_o.pc_set = 1'b1;
+          ctrl_fsm_o.pc_mux = PC_TRAP_CLICV_TGT;
+          ctrl_fsm_o.kill_if = 1'b1;
+          ctrl_fsm_o.kill_id = 1'b1; // No need to let this psuedo op continue todo: pass on to WB if exception
```
Silabs-ArjanB, line 754: Why not always pass it on? Also describe if there can for example be a instruction address breakpoint associated with this or not.

rtl/cv32e40x_core.sv
```diff
@@ -302,10 +307,6 @@ module cv32e40x_core import cv32e40x_pkg::*;
   assign irq_id  = ctrl_fsm.irq_id;
```
Silabs-ArjanB, line 307: Width mismatch

rtl/cv32e40x_core.sv
```diff
@@ -792,12 +800,31 @@ module cv32e40x_core import cv32e40x_pkg::*;
...
     if (SMCLIC) begin : gen_clic_interrupt
-      // Todo: instantiate cv32e40x_clic_int_controller
-      // For now just tie off outputs
-      assign irq_req_ctrl = '0;
-      assign irq_id_ctrl  = '0;
-      assign irq_wu_ctrl  = '0;
       assign mip          = '0;
+
+      cv32e40x_clic_int_controller
+      int_controller_i
```
Silabs-ArjanB, line 806: int_controller_i -> clic_int_controller_i

rtl/cv32e40x_if_stage.sv
```diff
@@ -260,7 +271,8 @@ module cv32e40x_if_stage import cv32e40x_pkg::*;
       // alignment buffer has a valid instruction
       if (if_valid_o && id_ready_i) begin
         if_id_pipe_o.instr_valid      <= 1'b1;
-        if_id_pipe_o.instr            <= instr_decompressed;
+        // Bypass comressed decoder when result is a pointer
+        if_id_pipe_o.instr            <= prefetch_is_ptr ? prefetch_instr : instr_decompressed;
```
Silabs-ArjanB, line 275: This mux already exists inside the comrpessed decoder. Remove it here and adapt the compressed decoder instead. Also make sure inside that decoder that is_compressed_o = 0 for pointers.

rtl/cv32e40x_prefetcher.sv
```diff
@@ -47,13 +47,15 @@ module cv32e40x_prefetcher
...
   output logic                     fetch_ready_o,
+  input  logic                     fetch_data_access_i,           // Access is data access (CLIC and Zce)
 
   // Transaction request interface
   output logic                     trans_valid_o,           // Transaction request valid (to bus interface adapter)
   input  logic                     trans_ready_i,           // Transaction request ready (transaction gets accepted when trans_valid_o and trans_ready_i are both 1)
-  output logic [31:0]              trans_addr_o             // Transaction address (only valid when trans_valid_o = 1). No stability requirements.
+  output logic [31:0]              trans_addr_o,            // Transaction address (only valid when trans_valid_o = 1). No stability requirements.
+  output logic                     trans_data_access_o      // Transaction is treated as a data access (CLIC and Zce)
+
```
Silabs-ArjanB, line 57: Zce -> Zc
Transaction is treated as a data access (1 for CLIC, 0 for Zc)

rtl/include/cv32e40x_pkg.sv
```diff
@@ -167,7 +167,7 @@ typedef enum logic [DIV_OP_WIDTH-1:0]
  } div_opcode_e;
 
 // FSM state encoding
-typedef enum logic [2:0] { RESET, BOOT_SET, FUNCTIONAL, SLEEP, DEBUG_TAKEN} ctrl_state_e;
+typedef enum logic [2:0] { RESET, BOOT_SET, FUNCTIONAL, SLEEP, DEBUG_TAKEN, VECTORING} ctrl_state_e;
```
Silabs-ArjanB, line 170: VECTORING -> POINTER_FETCH

rtl/cv32e40x_prefetcher.sv
```diff
@@ -74,6 +76,8 @@ module cv32e40x_prefetcher
 
   assign fetch_ready_o = trans_valid_o && trans_ready_i;
 
+  assign trans_data_access_o = fetch_data_access_i;
```
Silabs-ArjanB, line 79: I think that this assignment is the reason why the pointer fetch control is a bit awkward. trans_data_access_o assignments should be handled in a similar manner as the trans_addr_o assignments. Put an assign to this signal whereever there is an assign to trans_addr_o to make this easier to review.

rtl/include/cv32e40x_pkg.sv
```diff
@@ -833,7 +833,9 @@ typedef enum logic[3:0] {
   PC_TRAP_IRQ = 4'b1001,
   PC_TRAP_DBD = 4'b1010,
   PC_TRAP_DBE = 4'b1011,
-  PC_TRAP_NMI = 4'b1100
+  PC_TRAP_NMI = 4'b1100,
+  PC_TRAP_CLICV = 4'b1101,
```
Silabs-ArjanB, line 837: Will this get extended with more value when implementing table jump; if so, how? (I might want to rename these, but it depends on how table jump will be handled)

rtl/include/cv32e40x_pkg.sv
```diff
@@ -1006,6 +1008,7 @@ typedef struct packed {
 typedef struct packed
 {
   logic        compressed;
+  logic        clicv;
```
Silabs-ArjanB, line 1011: Similar, what will this look like after adding table jump?

rtl/include/cv32e40x_pkg.sv
```diff
@@ -1180,14 +1183,16 @@ typedef struct packed {
...
 
   // To WB stage
   logic        block_data_addr;       // To LSU to prevent data_addr_wb_i updates between error and taken NMI
-  logic [4:0]  m_exc_vec_pc_mux;      // id of taken irq (to IF, EXC_PC_MUX, zeroed if mtvec_mode==0)
+  logic [10:0] m_exc_vec_pc_mux;      // id of taken irq (to IF, EXC_PC_MUX, zeroed if mtvec_mode==0)
+                                      // Setting to 11 bits (max), unused bits will be tied off
 
   logic        irq_ack;               // irq has been taken
-  logic [4:0]  irq_id;                // id of taken irq
+  logic [10:0] irq_id;                // id of taken irq. Max width, unused bits will be tied off
```
Silabs-ArjanB, line 1195: There will be max 1024 interrupts, why 11 bits? If it is due to NMIs, why did 5 bits suffice before?

sva/cv32e40x_alignment_buffer_sva.sv
```diff
@@ -244,5 +246,17 @@ module cv32e40x_alignment_buffer_sva
       else
         `uvm_error("Alignment buffer SVA", "Read pointer(2) illegal value")
 
+  // Shall not fetch anything between pointer fetch and the actual instruction fetch
+  // based on the pointer.
+  property p_single_ptr_fetch;
+    @(posedge clk) disable iff (!rst_n)
+    (fetch_valid_o && fetch_ready_i && fetch_data_access_o) |=> !fetch_valid_o until ctrl_fsm_i.pc_set;
+  endproperty
```
Silabs-ArjanB, line 254: Nice!

sva/cv32e40x_alignment_buffer_sva.sv
```diff
@@ -244,5 +246,17 @@ module cv32e40x_alignment_buffer_sva
...
+
+  a_single_ptr_fetch:
+    assert property(p_single_ptr_fetch)
+    else
```
Silabs-ArjanB, line 258: Zce -> Zc

rtl/cv32e40x_if_stage.sv
```diff
@@ -169,40 +178,41 @@ module cv32e40x_if_stage import cv32e40x_pkg::*;
   assign core_trans.prot[2:1] = PRIV_LVL_M;             // Machine mode
   assign core_trans.memtype = 2'b00;                    // memtype is assigned in the MPU, tie off.
 
+  // todo: CLIC: Vector loads must respect mprv
```
Silabs-ArjanB, line 181: prot[0] assignment needs to be updated. 
Add assertion that no back-to-back data (pointer) fetches can occur on instruction OBI.

rtl/cv32e40x_alignment_buffer.sv
```diff
@@ -49,7 +50,8 @@ module cv32e40x_alignment_buffer import cv32e40x_pkg::*;
   output logic           instr_valid_o,
   input  logic           instr_ready_i,
   output inst_resp_t     instr_instr_o,
-  output logic [31:0]    instr_addr_o
+  output logic [31:0]    instr_addr_o,
```
Silabs-ArjanB, line 53: Add assertion that the address from where the pointer was fetched is word-aligned.

rtl/cv32e40x_alignment_buffer.sv
```diff
@@ -508,6 +517,7 @@ module cv32e40x_alignment_buffer import cv32e40x_pkg::*;
         // Reset pointers on branch
         wptr <= 'd0;
         rptr <= 'd0;
+        is_ptr_q <= ctrl_fsm_i.pc_set_clicv;
```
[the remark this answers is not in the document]
silabs-oysteink: Yes, I believe that one can be shared.

rtl/cv32e40x_alignment_buffer.sv
```diff
@@ -109,7 +116,7 @@ module cv32e40x_alignment_buffer import cv32e40x_pkg::*;
   // Request a transfer when needed, or we do a branch, iff outstanding_cnt_q is less than 2
   // Adjust instruction count with 'pop_q', which tells if we consumed an instruction
   // during the last cycle
-  assign fetch_valid_o = ctrl_fsm_i.instr_req &&
+  assign fetch_valid_o = (ctrl_fsm_i.instr_req || !ptr_fetch_done_q) &&
```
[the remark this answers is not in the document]
silabs-oysteink: Agreed, it makes the code a little less clean. One alternative would be to let the controller know the state of fetches being accepted or not, but I'm not sure if that is any cleaner.

rtl/cv32e40x_alignment_buffer.sv
```diff
@@ -472,17 +479,19 @@ module cv32e40x_alignment_buffer import cv32e40x_pkg::*;
...
+      instr_cnt_q       <= 'd0;
       outstanding_cnt_q <= 'd0;
-      rptr      <= 'd0;
-      wptr      <= 'd0;
-      pop_q     <= 1'b0;
+      rptr              <= 'd0;
+      wptr              <= 'd0;
+      pop_q             <= 1'b0;
+      is_ptr_q          <= 1'b0;
+      ptr_fetch_done_q  <= 1'b1; // Reset to 1 to avoid pointer fetches out of reset
```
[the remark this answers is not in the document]
silabs-oysteink: Comment added. We could of course flip the polarity of the bit, so it would become "ptr_fetch_pending_q" instead, causing a better reset value of 0.

rtl/cv32e40x_alignment_buffer.sv
silabs-oysteink: Fixed

rtl/cv32e40x_alignment_buffer.sv
```diff
@@ -535,4 +557,14 @@ module cv32e40x_alignment_buffer import cv32e40x_pkg::*;
...
+  // Signal that result is a pointer
+  // CLIC vectoring or Zce table jump
+  assign instr_is_ptr_o = is_ptr_q;
+
+  // Fetch is treated as a data access for CLIC vector fetch
+  // Only set for the initial pc_set_clicv point fetch
+  // If the pointer fetch isn't granted immediately, is_ptr_q will be true until
+  // the cycle after the next pc_set. Must avoid signaling fetch_data_access_o
+  // for the first pc_set after pointer fetch.
+  assign fetch_data_access_o = ctrl_fsm_i.pc_set_clicv || (is_ptr_q && !(ctrl_fsm_i.pc_set && !ctrl_fsm_i.pc_set_clicv));
```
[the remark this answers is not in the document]
silabs-oysteink: pc_set is implied, but for clarity we should add it anyway.

rtl/cv32e40x_clic_int_controller.sv
silabs-oysteink: Fixed

rtl/cv32e40x_controller_fsm.sv
silabs-oysteink: The MPU would need some more state, so I think it is cleaner to to do it in the controller.

rtl/cv32e40x_controller_fsm.sv
silabs-oysteink: Fixed

rtl/cv32e40x_controller_fsm.sv
silabs-oysteink: Fixed

rtl/cv32e40x_controller_fsm.sv
silabs-oysteink: Fixed

silabs-oysteink: Fixed

rtl/cv32e40x_core.sv
silabs-oysteink: Fixed

rtl/cv32e40x_controller_fsm.sv
```diff
@@ -515,15 +519,25 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
...
 
           ctrl_fsm_o.csr_save_cause  = 1'b1;
           ctrl_fsm_o.csr_cause.irq = 1'b1;
-          ctrl_fsm_o.csr_cause.exception_code = {3'b000, irq_id_ctrl_i};
+
+
+          if (SMCLIC) begin
+            ctrl_fsm_o.csr_cause.exception_code = {3'b000, irq_id_ctrl_i}; // todo: factor in SMCLIC_ID_WIDTH --
+            ctrl_fsm_o.pc_mux = PC_TRAP_CLICV;
+            ctrl_fsm_ns = VECTORING;
```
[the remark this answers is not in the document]
silabs-oysteink: Agreed. Renamed to PTR_FETCH.

rtl/cv32e40x_cs_registers.sv
```diff
@@ -974,7 +978,7 @@ module cv32e40x_cs_registers import cv32e40x_pkg::*;
       );
 
       assign mie_q  = 32'h0;
-
+      assign mtvt_addr_o = mtvt_q.addr[31:(32-MTVT_ADDR_WIDTH)];
```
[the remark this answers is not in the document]
silabs-oysteink: Done

rtl/include/cv32e40x_pkg.sv
silabs-oysteink: Fixed

sva/cv32e40x_alignment_buffer_sva.sv
silabs-oysteink: Fixed

rtl/cv32e40x_if_stage.sv
```diff
@@ -120,11 +125,13 @@ module cv32e40x_if_stage import cv32e40x_pkg::*;
...
       PC_FENCEI:   branch_addr_n = ctrl_fsm_i.pipe_pc;                                          // Jump to next instruction forces prefetch buffer reload
       PC_TRAP_EXC: branch_addr_n = {mtvec_addr_i, 7'h0};                                        // All the exceptions go to base address
-      PC_TRAP_IRQ: branch_addr_n = {mtvec_addr_i, ctrl_fsm_i.m_exc_vec_pc_mux, 2'b00};          // interrupts are vectored
+      PC_TRAP_IRQ: branch_addr_n = {mtvec_addr_i, ctrl_fsm_i.m_exc_vec_pc_mux[4:0], 2'b00};     // interrupts are vectored
       PC_TRAP_DBD: branch_addr_n = {dm_halt_addr_i[31:2], 2'b0};
       PC_TRAP_DBE: branch_addr_n = {dm_exception_addr_i[31:2], 2'b0};
       PC_TRAP_NMI: branch_addr_n = USE_DEPRECATED_FEATURE_SET ? {nmi_addr_i[31:2], 2'b00} :
                                                                 {mtvec_addr_i, NMI_MTVEC_INDEX, 2'b00};
+      PC_TRAP_CLICV: branch_addr_n = {mtvt_addr_i, ctrl_fsm_i.m_exc_vec_pc_mux, 2'b00}; // todo: use only relevant part of m_exc_vec_pc_mux
+      PC_TRAP_CLICV_TGT: branch_addr_n = if_id_pipe_o.instr.bus_resp.rdata;
```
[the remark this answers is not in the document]
silabs-oysteink: Fixed

rtl/cv32e40x_if_stage.sv
```diff
@@ -169,40 +178,41 @@ module cv32e40x_if_stage import cv32e40x_pkg::*;
...
-    .bus_resp_valid_i     ( bus_resp_valid          ),
-    .bus_resp_i           ( bus_resp                )
+    .clk                  ( clk                         ),
+    .rst_n                ( rst_n                       ),
+    .atomic_access_i      ( 1'b0                        ), // No atomic transfers on instruction side
+    .misaligned_access_i  ( 1'b0                        ), // MPU on instruction side will not issue misaligned access fault
+                                                           // Misaligned access to main is allowed, and accesses outside main will
+                                                           // result in instruction access fault (which will have priority over
+                                                           //  misaligned from I/O fault)
+    .if_data_access_i     ( prefetch_trans_data_access  ), // Indicate data access from IF stage. TODO: Use for table jumps (?)
```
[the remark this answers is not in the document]
silabs-oysteink: Fixed

rtl/cv32e40x_alignment_buffer.sv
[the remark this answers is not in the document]
Silabs-ArjanB: I expect that it will be cleanest if the controller simply does not lower instr_req for the pointer fetch scenario. I expect that can be a local decision inside the alignment buffer.

rtl/cv32e40x_controller_fsm.sv
Silabs-ArjanB: Agreed

rtl/cv32e40x_alignment_buffer.sv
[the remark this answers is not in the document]
silabs-oysteink: Agreed, I'll implement that.

rtl/cv32e40x_prefetcher.sv
silabs-oysteink: Fixed

rtl/cv32e40x_controller_fsm.sv
silabs-oysteink: Fixed

rtl/cv32e40x_controller_fsm.sv
silabs-oysteink: Now always passed on

rtl/cv32e40x_core.sv
silabs-oysteink: Fixed

rtl/cv32e40x_if_stage.sv
silabs-oysteink: Done

rtl/include/cv32e40x_pkg.sv
silabs-oysteink: Yes, I think we need another entry to keep the code clean, as the table jumps will reference the jvt CSR instead of the mtvt. PC_TRAP_TBLJMP with a TBD value.

rtl/include/cv32e40x_pkg.sv
silabs-oysteink: I guess we need a separate bit since they are treated differently with respect to exceptions. 'tbljmp' could be a candiate name.

rtl/include/cv32e40x_pkg.sv
silabs-oysteink: Wops, fixed.

sva/cv32e40x_rvfi_sva.sv
```diff
@@ -51,10 +51,12 @@ module cv32e40x_rvfi_sva
   // Ideally, we should assert that every irq_ack eventually leads to rvfi_intr,
   // but since there can be an infinite delay between irq_ack and rvfi_intr (e.g. because of bus stalls), we're settling for asserting
   // that irq_ack leads to RVFI capturing a trap (in_trap[IF_STAGE] = 1)
+  /* todo: adapt to CLIC
```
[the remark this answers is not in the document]
silabs-oysteink: I think we need to fix/adapt RVFI. I will look into it, it shouldn't be too much work.

rtl/cv32e40x_alignment_buffer.sv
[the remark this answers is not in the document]
silabs-oysteink: Fixed/restructured code.

rtl/cv32e40x_alignment_buffer.sv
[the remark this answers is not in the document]
silabs-oysteink: Fixed/restructured.

rtl/cv32e40x_clic_int_controller.sv
silabs-oysteink: Now resetting to machine mode. The assertion would probably better be placed outside of the core, as the core external part of CLIC is driving this signal.

rtl/cv32e40x_if_stage.sv
```diff
@@ -237,10 +247,11 @@ module cv32e40x_if_stage import cv32e40x_pkg::*;
   assign if_busy_o = prefetch_busy;
 
   // Populate instruction meta data
-  instr_meta_t instr_meta_n; 
+  instr_meta_t instr_meta_n;
   always_comb begin
     instr_meta_n = '0;
     instr_meta_n.compressed = instr_compressed_int;
+    instr_meta_n.clicv      = prefetch_is_ptr;
```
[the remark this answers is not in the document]
silabs-oysteink: I think it will be cleaner if we have two different flags, 'clic_ptr' and 'tbljmp_ptr' for instance? They will likely be treated a bit different when it comes to exceptions.

rtl/cv32e40x_prefetcher.sv
silabs-oysteink: Done

rtl/cv32e40x_if_stage.sv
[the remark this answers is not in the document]
Silabs-ArjanB: sounds good

sva/cv32e40x_decoder_sva.sv
```diff
@@ -50,15 +50,16 @@ module cv32e40x_decoder_sva
...
   // Check that the two LSB of the incoming instructions word is always 2'b11
   // Predecoder should always emit uncompressed instructions
   property p_uncompressed_lsb;
     @(posedge clk) disable iff(!rst_n)
       (instr_rdata_i[1:0] == 2'b11);
   endproperty
 
-  a_uncompressed_lsb: assert property(p_uncompressed_lsb) else `uvm_error("decoder", "2 LSBs not 2'b11")
+  // todo: adapt to CLIC addresses in instr field
+  //a_uncompressed_lsb: assert property(p_uncompressed_lsb) else `uvm_error("decoder", "2 LSBs not 2'b11")
```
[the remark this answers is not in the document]
silabs-oysteink: Done

rtl/cv32e40x_alignment_buffer.sv
```diff
@@ -521,6 +531,18 @@ module cv32e40x_alignment_buffer import cv32e40x_pkg::*;
         end
       end
 
+      // Set pointer fetch done flag when a fetch is granted
+      // If the fetch is accepted the same cycle as the pointer request,
+      // the flag will remain high.
+      if(fetch_valid_o && fetch_ready_i) begin
+        ptr_fetch_done_q <= 1'b1;
```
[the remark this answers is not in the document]
silabs-oysteink: Semantics changed, should be more intuitive now.

rtl/cv32e40x_alignment_buffer.sv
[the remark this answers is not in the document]
silabs-oysteink: Moved to prefetcher

rtl/cv32e40x_alignment_buffer.sv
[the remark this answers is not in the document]
silabs-oysteink: Fixed

rtl/cv32e40x_alignment_buffer.sv
silabs-oysteink: This already exists in the prefetcher_sva (p_fetch_branch_addr_aligned)

rtl/cv32e40x_if_stage.sv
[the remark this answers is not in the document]
silabs-oysteink: Renamed to meta.clic_ptr

rtl/cv32e40x_cs_registers.sv
```diff
@@ -724,8 +742,19 @@ module cv32e40x_cs_registers import cv32e40x_pkg::*;
             mepc_n  = ctrl_fsm_i.pipe_pc;
             mepc_we = 1'b1;
 
+            // Set mcause from controller
             mcause_n  = ctrl_fsm_i.csr_cause;
+
+            // mpil is saved from mintstatus
+            mcause_n.mpil = mintstatus_q.mil;
```
Silabs-ArjanB, line 749: This needs an explanation on why this works if SMCLIC = 0. Even better would be to explicitly set it to 0 in that case.

Maybe best to guard all CLIC related assignments in this file with SMCLIC and use 0 assignments if SMCLIC = 0.

rtl/cv32e40x_cs_registers.sv
```diff
@@ -724,8 +742,19 @@ module cv32e40x_cs_registers import cv32e40x_pkg::*;
...
             mcause_we = 1'b1;
+
+            // todo: handle exception vs interrupt
+            // Save new interrupt level to mintstatus
+            mintstatus_n.mil = ctrl_fsm_i.irq_level;
```
Silabs-ArjanB, line 754: This register does not exist when SMCLIC = 0 I would prefer if we can exclude such assignments (also following line) if SMCLIC = 0.

rtl/cv32e40x_cs_registers.sv
```diff
@@ -734,8 +763,18 @@ module cv32e40x_cs_registers import cv32e40x_pkg::*;
...
         mstatus_n.mpp  = PRIV_LVL_M;
         mstatus_we = 1'b1;
+
+        mintstatus_n.mil = mcause_q.mpil;
+        mintstatus_we = 1'b1;
       end //ctrl_fsm_i.csr_restore_mret
 
+      // mcause.minhv shall be cleared if vector fetch is successful
+      ctrl_fsm_i.csr_clear_minhv: begin
+        mcause_n = mcause_q;
```
Silabs-ArjanB, line 773: Guard all content in this block. Should not get triggered if SMCLIC = 0.

rtl/cv32e40x_if_stage.sv
```diff
@@ -260,6 +274,7 @@ module cv32e40x_if_stage import cv32e40x_pkg::*;
       // alignment buffer has a valid instruction
       if (if_valid_o && id_ready_i) begin
         if_id_pipe_o.instr_valid      <= 1'b1;
+        // Bypass comressed decoder when result is a pointer
```
Silabs-ArjanB, line 277: Update comment (pointer is passed on via compressed decoder)

rtl/cv32e40x_controller_fsm.sv
```diff
@@ -193,10 +197,15 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
   // Once the fencei handshake is initiated, it must complete and the instruction must retire.
   // The instruction retires when fencei_req_and_ack_q = 1
   assign fencei_ongoing = fencei_flush_req_o || fencei_req_and_ack_q;
-  
+
   // Mux selector for vectored IRQ PC
-  assign ctrl_fsm_o.m_exc_vec_pc_mux = (mtvec_mode_i == 2'b0) ? 5'h0 : exc_cause;
-  
+  // Used for both basic mode and CLIC when shv == 0.
+  assign ctrl_fsm_o.mtvec_pc_mux = ((mtvec_mode_i == 2'b0) || (mtvec_mode_i == 2'b11 && !irq_clic_shv_i)) ? 5'h0 : exc_cause[4:0];
```
Silabs-ArjanB, line 203: brakets around the == comparison

rtl/cv32e40x_controller_fsm.sv
```diff
@@ -212,29 +221,33 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
...
 
   // Exception in WB if the following evaluates to 1
-  assign exception_in_wb = ((ex_wb_pipe_i.instr.mpu_status != MPU_OK)            ||
-                            ex_wb_pipe_i.instr.bus_resp.err                      ||
-                            ex_wb_pipe_i.illegal_insn                            ||
-                            (ex_wb_pipe_i.sys_en && ex_wb_pipe_i.sys_ecall_insn) ||
-                            (ex_wb_pipe_i.sys_en && ex_wb_pipe_i.sys_ebrk_insn)  ||
+  // CLIC: bus errors for pointer fetches are treated as NMI, not exceptions.
+  assign exception_in_wb = ((ex_wb_pipe_i.instr.mpu_status != MPU_OK)                              ||
+                            !ex_wb_pipe_i.instr_meta.clic_ptr && ex_wb_pipe_i.instr.bus_resp.err   ||
```
Silabs-ArjanB, line 235: brackets

rtl/cv32e40x_controller_fsm.sv
```diff
@@ -212,29 +221,33 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
...
   // Set exception cause
-  assign exception_cause_wb = ex_wb_pipe_i.instr.mpu_status != MPU_OK              ? EXC_CAUSE_INSTR_FAULT     :
-                              ex_wb_pipe_i.instr.bus_resp.err                      ? EXC_CAUSE_INSTR_BUS_FAULT :
-                              ex_wb_pipe_i.illegal_insn                            ? EXC_CAUSE_ILLEGAL_INSN    :
-                              (ex_wb_pipe_i.sys_en && ex_wb_pipe_i.sys_ecall_insn) ? EXC_CAUSE_ECALL_MMODE     :
-                              (ex_wb_pipe_i.sys_en && ex_wb_pipe_i.sys_ebrk_insn)  ? EXC_CAUSE_BREAKPOINT      :
-                              (lsu_mpu_status_wb_i == MPU_WR_FAULT)                ? EXC_CAUSE_STORE_FAULT     :
+  // For CLIC: Pointer fetches with PMA/PMP errors will get the exception code converted to LOAD_FAULT
+  //           Bus errors will be converted to NMI as for regular loads.
+  assign exception_cause_wb = !ex_wb_pipe_i.instr_meta.clic_ptr && ex_wb_pipe_i.instr.mpu_status != MPU_OK   ? EXC_CAUSE_INSTR_FAULT     :
```
Silabs-ArjanB, line 244: Brackets around expressions used in select parts

rtl/cv32e40x_controller_fsm.sv
```diff
@@ -271,7 +284,10 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
   // Using flopped version to avoid paths from data_err_i/data_rvalid_i to instr_* outputs
   // Gating the pending signal instead of the allowed signal for debug related conditions, otherwise a pending NMI during debug mode
   // or single stepping with dcsr.stepie==0 would stall ID stage and we would never get out of debug, resulting in a deadlock.
-  assign pending_nmi = nmi_pending_q && !debug_mode_q && !(dcsr_i.step && !dcsr_i.stepie);
+  // CLIC pointer fetches with associated bus errors are treated as NMI. The error bit is taken from ex_wb_pipe and not flopped further.
+  //      This preserves the address of the pointer (ex_wb_pipe.pc) that must be stored to mepc.
+  assign pending_nmi = (nmi_pending_q || ex_wb_pipe_i.instr_valid && ex_wb_pipe_i.instr_meta.clic_ptr && ex_wb_pipe_i.instr.bus_resp.err) &&
```
Silabs-ArjanB, line 289: Brackets

Silabs-ArjanB: It seems you are not explicitly using nmi_pending_q and nmi_is_store_q. How will this NMI be treated correctly on rvfi_nmip?

rtl/cv32e40x_controller_fsm.sv
```diff
@@ -724,6 +759,25 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
...
         ctrl_fsm_ns = FUNCTIONAL;
       end
+      // State for CLIC vectoring (and Zc table jumps)
+      // In this state a fetch has been ordered, and the controller
+      // is waiting for the pointer to arrive in the decode stage.
+      POINTER_FETCH: begin
+        if (if_id_pipe_i.instr_meta.clic_ptr && if_id_pipe_i.instr_valid) begin
+          // Function pointer reached ID stage, do another jump
+          // if no faults happened during pointer fetch. (mcause.minhv will stay high for faults)
+          if(!((if_id_pipe_i.instr.mpu_status != MPU_OK) || if_id_pipe_i.instr.bus_resp.err)) begin
```
Silabs-ArjanB, line 769: Add todo to deal with integrity related faults for 40S

rtl/cv32e40x_controller_fsm.sv
```diff
@@ -724,6 +759,25 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
...
+            ctrl_fsm_o.pc_set = 1'b1;
+            ctrl_fsm_o.pc_mux = PC_TRAP_CLICV_TGT;
+            ctrl_fsm_o.kill_if = 1'b1;
+            ctrl_fsm_o.csr_clear_minhv = 1'b1;
+          end
+          // Note: If the pointer fetch faulted (pma/pmp/bus error), an exception or NMI will
+          // be taken once the pointer fetch reachces WB (two cycles after the current)
+          // The FSM must be in the FUNCTIONAL state to take the exception or NMI.
+          ctrl_fsm_ns = FUNCTIONAL;
```
Silabs-ArjanB, line 778: I think there might be a bug here. What happens if a ptr is in EX or WB and it is killed by an interrupt. What will the mepc be and what will happen upon re-execution from that mepc (will the ptr access be retried as an instruction fetch?). If this goes wrong (which I would expect) maybe the easiest solution is to disallow interrupts when there is a ptr in EX or WB. Additional question: What type of instruction can be in WB when there is a ptr in EX?

rtl/cv32e40x_cs_registers.sv
silabs-oysteink: Fixed

rtl/cv32e40x_cs_registers.sv
silabs-oysteink: Fixed

rtl/cv32e40x_cs_registers.sv
```diff
@@ -734,8 +763,18 @@ module cv32e40x_cs_registers import cv32e40x_pkg::*;
         mstatus_n.mpie = 1'b1;
         mstatus_n.mpp  = PRIV_LVL_M;
         mstatus_we = 1'b1;
+
+        mintstatus_n.mil = mcause_q.mpil;
```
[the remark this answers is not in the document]
silabs-oysteink: Fixed

rtl/cv32e40x_cs_registers.sv
silabs-oysteink: Fixed

rtl/cv32e40x_prefetcher.sv
```diff
@@ -104,6 +115,7 @@ module cv32e40x_prefetcher
         // occur if for example an interrupt is taken right after a taken jump which did not
         // yet have its target address accepted by the bus interface adapter.
         trans_addr_o = fetch_branch_i ? fetch_branch_addr_i : trans_addr_q;
+        trans_data_access_o = fetch_branch_i ? fetch_data_access_i : trans_data_access_q; // No new pointer fetch should be available in this case
```
Silabs-ArjanB, line 118: Can't this just be: trans_data_access_o = trans_data_access_q;

I would expect that if trans_data_access_q = 1 in this state, then fetch_branch_i must be 0. Can you confirm and add an assertion for that?

rtl/cv32e40x_clic_int_controller.sv
```diff
@@ -0,0 +1,137 @@
...
+  ///////////////////////////
+  // Outputs to controller //
+  ///////////////////////////
+
+  // Request to take interrupt if:
+  // There a pending-and-enabled interrupt and interrupts are enabled globally
+  // AND the incoming irq level is above the cores current effective interrupt level.
+  // todo: In user mode, machine threshold should not mask interrupts to machine mode
+  assign irq_req_ctrl_o = clic_irq_q &&
+                          clic_irq_level_q > effective_irq_level &&
```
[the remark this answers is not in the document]
silabs-oysteink: Fixed

rtl/cv32e40x_clic_int_controller.sv
```diff
@@ -0,0 +1,137 @@
...
+  // todo: move logic for m_ie_i from cs_registers to here.
+  assign global_irq_enable = m_ie_i;
+
+  assign effective_irq_level = (mintthresh_i > mintstatus_i.mil) ? mintthresh_i : mintstatus_i.mil;
+  ///////////////////////////
+  // Outputs to controller //
+  ///////////////////////////
+
+  // Request to take interrupt if:
+  // There a pending-and-enabled interrupt and interrupts are enabled globally
```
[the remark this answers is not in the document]
silabs-oysteink: Fixed

rtl/cv32e40x_clic_int_controller.sv
```diff
@@ -0,0 +1,137 @@
...
+  // AND the incoming irq level is above the cores current effective interrupt level.
```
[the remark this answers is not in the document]
silabs-oysteink: Fixed

rtl/cv32e40x_prefetcher.sv
```diff
@@ -123,12 +135,14 @@ module cv32e40x_prefetcher
...
       state_q        <= IDLE;
       trans_addr_q   <= '0;
+      trans_data_access_q <= 1'b0;
     end
     else
     begin
       state_q        <= next_state;
       if (fetch_branch_i || (trans_valid_o && trans_ready_i)) begin
         trans_addr_q <= trans_addr_o;
+        trans_data_access_q <= fetch_data_access_i;
```
Silabs-ArjanB, line 145: This looks inconsistent. Should't it be: trans_data_access_q <= fetch_data_access_o;

rtl/cv32e40x_if_stage.sv
silabs-oysteink: Fixed

rtl/cv32e40x_controller_fsm.sv
silabs-oysteink: Fixed

rtl/cv32e40x_controller_fsm.sv
silabs-oysteink: Fixed

rtl/cv32e40x_controller_fsm.sv
silabs-oysteink: Fixed

rtl/cv32e40x_wb_stage.sv
```diff
@@ -128,11 +128,11 @@ module cv32e40x_wb_stage import cv32e40x_pkg::*;
   //   cannot be used to increment the minstret CSR (as that should not increment for e.g. ecall, ebreak, etc.)
   // - Will be 1 only for the second phase of a split misaligned load/store that completes without MPU errors.
   //   If an MPU error occurs, wb_valid will be 1 due to lsu_exception (for any phase where the error occurs)
-
+  // - Will be 0 for CLIC pointer fetches todo: Do we need wb_valid=1 for faulted pointer fetches for RVFI?
   assign wb_valid = ((!ex_wb_pipe_i.lsu_en && !xif_waiting) ||    // Non-LSU instructions have valid result in WB, also for exceptions, unless we are waiting for a coprocessor
                      ( ex_wb_pipe_i.lsu_en && lsu_valid_i)  ||    // LSU instructions have valid result based on data_rvalid_i
                      ( ex_wb_pipe_i.lsu_en && lsu_exception)      // LSU instruction had an exception
-                    ) && instr_valid;
+                    ) && !ex_wb_pipe_i.instr_meta.clic_ptr && instr_valid;
```
Silabs-ArjanB, line 135: Why are ptr fetches excluded here? Isn't that changing the semantics of wb_valid?

rtl/include/cv32e40x_pkg.sv
```diff
@@ -1180,14 +1189,18 @@ typedef struct packed {
...
   logic        instr_req;             // Start fetching instructions
   logic        pc_set;                // jump to address set by pc_mux
+  logic        pc_set_clicv;          // Signal pc_set it for CLIC vectoring pointer load
   pc_mux_e     pc_mux;                // Selector in the Fetch stage to select the rigth PC (normal, jump ...)
 
   // To WB stage
   logic        block_data_addr;       // To LSU to prevent data_addr_wb_i updates between error and taken NMI
-  logic [4:0]  m_exc_vec_pc_mux;      // id of taken irq (to IF, EXC_PC_MUX, zeroed if mtvec_mode==0)
+  logic [4:0]  mtvec_pc_mux;          // id of taken basic mode irq (to IF, EXC_PC_MUX, zeroed if mtvec_mode==0)
+  logic [10:0] mtvt_pc_mux;           // id of taken CLIC irq (to IF, EXC_PC_MUX, zeroed if not shv)
```
Silabs-ArjanB, line 1198: This should be 10 bits

rtl/cv32e40x_controller_fsm.sv
```diff
@@ -193,10 +197,15 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
...
-  
+
   // Mux selector for vectored IRQ PC
-  assign ctrl_fsm_o.m_exc_vec_pc_mux = (mtvec_mode_i == 2'b0) ? 5'h0 : exc_cause;
-  
+  // Used for both basic mode and CLIC when shv == 0.
+  assign ctrl_fsm_o.mtvec_pc_mux = ((mtvec_mode_i == 2'b0) || (mtvec_mode_i == 2'b11 && !irq_clic_shv_i)) ? 5'h0 : exc_cause[4:0];
+
+  // CLIC mode vectored PC mux is always the same as exc_cause.
+  assign ctrl_fsm_o.mtvt_pc_mux = exc_cause;
```
Silabs-ArjanB, line 206: This should be lower 10 bits of exc_cause only

Silabs-ArjanB (changes requested):

rtl/cv32e40x_controller_fsm.sv
```diff
@@ -32,7 +32,9 @@
 module cv32e40x_controller_fsm import cv32e40x_pkg::*;
 #(
   parameter bit       USE_DEPRECATED_FEATURE_SET = 1, // todo: remove once related features are supported by iss
-  parameter bit       X_EXT           = 0
+  parameter bit       X_EXT           = 0,
+  parameter bit       SMCLIC          = 0,
```
Silabs-ArjanB, line 36: We should make sure that the write buffer has been flushed before a CLIC ptr fetch is attempted.

Silabs-ArjanB (changes requested):

rtl/cv32e40x_mpu.sv
```diff
@@ -37,7 +37,7 @@ module cv32e40x_mpu import cv32e40x_pkg::*;
 
    input logic  atomic_access_i,     // Indicate that ongoing access is atomic
    input logic  misaligned_access_i, // Indicate that ongoing access is part of a misaligned access
-   input logic  if_data_access_i,    // Indicate that ongoing access is a data access from the IF stage
+   input logic  core_data_access_i,    // Indicate that ongoing access is a data access from the IF stage
```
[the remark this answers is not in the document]
silabs-oysteink: Fixed

rtl/cv32e40x_prefetcher.sv
silabs-oysteink: I agree, fixed.

rtl/include/cv32e40x_pkg.sv
silabs-oysteink: Fixed

sva/cv32e40x_decoder_sva.sv
```diff
@@ -50,14 +51,16 @@ module cv32e40x_decoder_sva
...
   a_b_dec_idle : assert property(p_idle_dec(decoder_b_ctrl)) else `uvm_error("decoder", "Assertion a_b_dec_idle failed")
-  
+
   // Check that the two LSB of the incoming instructions word is always 2'b11
   // Predecoder should always emit uncompressed instructions
+  // Exclude CLIC pointers
   property p_uncompressed_lsb;
     @(posedge clk) disable iff(!rst_n)
-      (instr_rdata_i[1:0] == 2'b11);
+      !if_id_pipe.instr_meta.clic_ptr |-> (instr_rdata_i[1:0] == 2'b11);
```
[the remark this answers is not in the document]
silabs-oysteink: Done

rtl/cv32e40x_controller_fsm.sv
silabs-oysteink: Fixed

rtl/cv32e40x_controller_fsm.sv
silabs-oysteink: Well spotted, I will look into this.

rtl/cv32e40x_controller_fsm.sv
silabs-oysteink: For CLIC pointers, I believe it is not possible to take a new interrupt when the pointer is still in the pipeline. We disable global interrupts in mstatus when taking starting the pointer fetch, and this needs to be re-enabled before we can take a new interrupt. No new instructions will complete before the pointer is out of the pipeline. We may however be interrupted by an external debug request, this could possibly cause some issues (I will have to dig a bit into this case). The PC of the pipeline stage containing the pointer would be the address of the pointer itself, so in theory we could be able to restart the pointer fetch - but more logic would be needed. Perhaps the 'lsu_interruptible' signal should only be allowed to go high when the write buffer is also empty?

When a (CLIC) pointer is in EX, there should be no valid instruction in WB, as the whole pipeline is killed when taking the interrupt and starting the pointer fetch. For Zc tablejumps this may not be the case though.

rtl/cv32e40x_cs_registers.sv
```diff
@@ -976,6 +1019,18 @@ module cv32e40x_cs_registers import cv32e40x_pkg::*;
       assign mie_q  = 32'h0;
 
     end else begin
+      cv32e40x_csr #(
```
Silabs-ArjanB, line 1022: Need block scope name on line above

Silabs-ArjanB (changes requested):

rtl/cv32e40x_cs_registers.sv
```diff
@@ -584,23 +590,30 @@ module cv32e40x_cs_registers import cv32e40x_pkg::*;
...
     mintthresh_we            = 1'b0;
-    mscratchcsw_n            = '0;
+    mscratchcsw_n            = csr_wdata_int;
     mscratchcsw_we           = 1'b0;
-    mscratchcswl_n           = '0;
+    mscratchcswl_n           = csr_wdata_int;
     mscratchcswl_we          = 1'b0;
     mie_n                    = csr_wdata_int & IRQ_MASK;
     mie_we                   = 1'b0;
+    mclicbase_n              = {csr_wdata_int[31:12], 12'h000};
```
Silabs-ArjanB, line 615: Also make it obvious that signals as mie_n (and similar others if existing) are 0 when SMCLIC = 1 (I know these are not used at the moment when SMCLIC = 1, but it is still good to make it explicit)

Silabs-ArjanB (changes requested):

rtl/cv32e40x_prefetcher.sv
silabs-oysteink: This should be true, I'll add the assertion and confirm.

rtl/cv32e40x_cs_registers.sv
silabs-oysteink: Fixed

rtl/cv32e40x_controller_fsm.sv
```diff
@@ -639,7 +674,7 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
           end
 
           // Mret in WB restores CSR regs
-          // 
+          //
```
Silabs-ArjanB, line 677: MRET is not implemented according to CLIC spec when SMCLIC=1 (see page 20 of CLIC spec explaining the dependence on mcause.inhv)

Silabs-ArjanB (changes requested):

rtl/cv32e40x_cs_registers.sv
```diff
@@ -584,23 +590,30 @@ module cv32e40x_cs_registers import cv32e40x_pkg::*;
       mtvec_n.addr[7]        = 1'b0; // todo : remove
     end
     mtvec_n.zero0            = mtvec_q.zero0;
-    mtvec_n.mode             = csr_mtvec_init_i ? mtvec_q.mode : {1'b0, csr_wdata_int[0]};
+
```
Silabs-ArjanB, line 593: mcause's minhv and mpil bitfields should be writeable by software when SMCLIC = 1. Our user manual also wrongly states minhv as a R bitfield (I will fix the documentation part)

Silabs-ArjanB (changes requested):

rtl/cv32e40x_cs_registers.sv
```diff
@@ -584,23 +590,30 @@ module cv32e40x_cs_registers import cv32e40x_pkg::*;
...
+    mtvt_n                   = {csr_wdata_int[31:(32-MTVT_ADDR_WIDTH)], {(32-MTVT_ADDR_WIDTH){1'b0}}};
     mtvt_we                  = 1'b0;
-    mnxti_n                  = '0;
+    mnxti_n                  = '0; // todo: implement
     mnxti_we                 = 1'b0;
-    mintstatus_n             = MINTSTATUS_RESET_VAL;
+    mintstatus_n             = mintstatus_q; // All fields of mintstatus are read only
     mintstatus_we            = 1'b0;
-    mintthresh_n             = '0;
+    mintthresh_n             = csr_wdata_int[7:0];
```
Silabs-ArjanB, line 607: Width mismatch. Explicitly 0 top 24 bits.

Silabs-ArjanB (changes requested):

rtl/cv32e40x_cs_registers.sv
silabs-oysteink: Fixed

sva/cv32e40x_rvfi_sva.sv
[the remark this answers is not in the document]
silabs-oysteink: I guarded the assertion with !SMCLIC and added a todo as this needs a bit more work.

Added issue #494

rtl/cv32e40x_controller_fsm.sv
silabs-oysteink: Fixed

rtl/cv32e40x_prefetcher.sv
silabs-oysteink: I added the suggested assert, but is fails during single stepping into an interrupt handler. I added issue #496 to address this.

rtl/cv32e40x_wb_stage.sv
silabs-oysteink: I added issue #497 for this. Issue also addresses that we only signal wb_valid for the last part of a split misaligned LSU instruction.

rtl/cv32e40x_cs_registers.sv
silabs-oysteink: Fixed

rtl/cv32e40x_controller_fsm.sv
silabs-oysteink: Added issue #498 for this

rtl/cv32e40x_cs_registers.sv
silabs-oysteink: Fixed

rtl/cv32e40x_cs_registers.sv
```diff
@@ -584,23 +590,30 @@ module cv32e40x_cs_registers import cv32e40x_pkg::*;
       mtvec_n.addr[7]        = 1'b0; // todo : remove
     end
     mtvec_n.zero0            = mtvec_q.zero0;
-    mtvec_n.mode             = csr_mtvec_init_i ? mtvec_q.mode : {1'b0, csr_wdata_int[0]};
+
+    if (SMCLIC) begin
+      mtvec_n.mode             = mtvec_q.mode; // mode is WARL 0x3 when using CLIC
+    end else begin
+      mtvec_n.mode             = csr_mtvec_init_i ? mtvec_q.mode : {1'b0, csr_wdata_int[0]};
```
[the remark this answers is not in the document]
Silabs-ArjanB: Moved into https://github.com/openhwgroup/cv32e40x/issues/501

rtl/cv32e40x_controller_fsm.sv
silabs-oysteink: I think at some point we talked about reverting the 'previous_exception' part to instead set alu_en for instructions with exceptions to make ex_valid go high, we could probably use that same method for pointers?

silabs-oysteink:
With SMCLIC==0: SEC clean and passes ci_check (using modified core-v-verif adapted to interface changes for clic)

Silabs-ArjanB (approved):

PR #505, opened 2022-04-07
minstret rvfi reporting fix for WFI

Silabs-ArjanB:


silabs-oysteink (approved):
Passes hello-world with ISS enabled.

PR #506, opened 2022-04-08
Implementation of mnxti

silabs-oysteink:
MNXTI functionality should now be in.
Tested basic functionality of reading correct pointer addresses for pending interrupts when changing levels in the memory mapped registers.

SEC clean for SMCLIC=0

### DIFF
--- a/rtl/cv32e40x_core.sv
+++ b/rtl/cv32e40x_core.sv
@@ -202,10 +202,10 @@ module cv32e40x_core import cv32e40x_pkg::*;
   logic [31:0] csr_rdata;
   logic csr_counter_read;
 
-  // CLIC signals for returning pointer to handles
+  // CLIC signals for returning pointer addresses
   // when mnxti is accessed
-  logic        csr_clic_pa_valid;
-  logic [31:0] csr_clic_pa;
+  logic        csr_clic_pa_valid;   // A CSR access to mnxti has a valid ponter address
+  logic [31:0] csr_clic_pa;         // Pointer address returned by accessing mnxti
 
   // LSU
   logic        lsu_split_ex;
@@ -276,6 +276,7 @@ module cv32e40x_core import cv32e40x_pkg::*;
   logic [7:0]                 irq_clic_level;
   logic                       mnxti_irq_pending;
   logic [SMCLIC_ID_WIDTH-1:0] mnxti_irq_id;
+  logic [7:0]                 mnxti_irq_level;
 
   // Used (only) by verification environment
   logic        irq_ack;
@@ -715,6 +716,7 @@ module cv32e40x_core import cv32e40x_pkg::*;
     .mcause_o                   ( mcause                 ),
     .mnxti_irq_pending_i        ( mnxti_irq_pending      ),
     .mnxti_irq_id_i             ( mnxti_irq_id           ),
+    .mnxti_irq_level_i          ( mnxti_irq_level        ),
     .clic_pa_valid_o            ( csr_clic_pa_valid      ),
     .clic_pa_o                  ( csr_clic_pa            ),
 
@@ -867,7 +869,8 @@ module cv32e40x_core import cv32e40x_pkg::*;
 
         // To cv32e40x_cs_registers
         .mnxti_irq_pending_o  ( mnxti_irq_pending  ),
-        .mnxti_irq_id_o       ( mnxti_irq_id       )
+        .mnxti_irq_id_o       ( mnxti_irq_id       ),
+        .mnxti_irq_level_o    ( mnxti_irq_level    )
       );
     end else begin : gen_basic_interrupt
       cv32e40x_int_controller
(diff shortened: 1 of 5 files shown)


rtl/cv32e40x_core.sv
```diff
@@ -197,9 +197,16 @@ module cv32e40x_core import cv32e40x_pkg::*;
...
   mintstatus_t mintstatus;
 
+  mcause_t     mcause;
+
   logic [31:0] csr_rdata;
   logic csr_counter_read;
 
+  // CLIC signals for returning pointer to handles
+  // when mnxti is accessed
+  logic        csr_clic_pa_valid;
```
Silabs-ArjanB, line 207: What does 'pa' stand for?

rtl/cv32e40x_cs_registers.sv
```diff
@@ -752,6 +763,50 @@ module cv32e40x_cs_registers import cv32e40x_pkg::*;
...
+    if (SMCLIC) begin
+      // In CLIC mode, writes to mcause.mpp/mpie is aliased to mstatus.mpp/mpie
+      // All other mstatus bits are preserved
+      if (mcause_we) begin
+        mstatus_n      = mstatus_q; // Preserve all fields
+
+        // Write mpie and mpp as aliased through mcause
+        mstatus_n.mpie = csr_wdata_int[MCAUSE_MPIE_BIT];
+        mstatus_n.mpp  = PRIV_LVL_M; // todo: handle priv mode for E40S
+      end else if (mnxti_we) begin
```
Silabs-ArjanB, line 779: If there is a mnxti write with side effects, then both mcause_we = 1 and mnxti_we = 1 and this block of code will not get executed.

rtl/cv32e40x_cs_registers.sv
```diff
@@ -752,6 +763,50 @@ module cv32e40x_cs_registers import cv32e40x_pkg::*;
...
+        // A mnxti write writes to mstatus.mie
+        mstatus_n     = mstatus_q;
```
Silabs-ArjanB, line 781: I don't think these two lines are correct. The spec says 'All CSR instructions may be used with xnxti. The operation will be the same as the corresponding CSR instruction used with xstatus.', so shouldn't the already computed mstatus_n from line 571 do the trick?

rtl/cv32e40x_cs_registers.sv
```diff
@@ -752,6 +763,50 @@ module cv32e40x_cs_registers import cv32e40x_pkg::*;
...
+          mcause_n.irq = 1'b1;
+          mcause_n.exception_code = {1'b0, 10'(mnxti_irq_id_i)};
+        end
+      end
+
+      // The CLIC pointer address should always be output for an access to MNXTI,
+      // but will only contain a nonzero value if a CLIC interrupt is actually pending
+      // with a higher level. The valid below will be high also for the cases where
+      // no side effects occur.
+      clic_pa_valid_o = csr_en_gated && (csr_waddr == CSR_MNXTI);
```
Silabs-ArjanB, line 801: Would the following not be more logical?

clic_pa_valid_o = csr_we_int && (csr_waddr == CSR_MNXTI);

Silabs-ArjanB (changes requested):
(We will discuss https://github.com/openhwgroup/cv32e40x/issues/507 separately as that issue was not introduced with this PR)

rtl/cv32e40x_core.sv
silabs-oysteink: PointerAddress, I tried to make a short name, but this is probably too short. Would 'csr_clic_ptr_addr_valid' be better perhaps?

rtl/cv32e40x_cs_registers.sv
silabs-oysteink: That's right, we'll have to check the state for both write enables in the first if block I think.

rtl/cv32e40x_cs_registers.sv
silabs-oysteink: The address for the pointer should be returned even for pure reads, so checking using we_int would break the spec for 'csrr' for instance

rtl/cv32e40x_core.sv
Silabs-ArjanB: You can keep this name; just add a comment on the lines where these signals are declared that contains at least the words 'pointer address'

rtl/cv32e40x_cs_registers.sv
Silabs-ArjanB: Maybe you can handle mnxti_we in the top/first if to get it correct?

rtl/cv32e40x_cs_registers.sv
Silabs-ArjanB: Yes, okay, thanks

rtl/cv32e40x_cs_registers.sv
silabs-oysteink: Fixed

rtl/cv32e40x_core.sv
silabs-oysteink: Fixed

rtl/cv32e40x_cs_registers.sv
silabs-oysteink: Agreed, fixed.

Silabs-ArjanB (approved):

PR #511, opened 2022-04-20
Bugfix: mnxti data forwarding causing wrong operand values

silabs-oysteink:
Fixed bug where a results forwarded from EX due to a mnxti access caused wrong operand values

as the data forwareded is not the data written to the register file.


rtl/cv32e40x_controller_bypass.sv
```diff
@@ -129,6 +129,12 @@ module cv32e40x_controller_bypass import cv32e40x_pkg::*;
   // Prevent load/store following a WFI in the pipeline
   assign ctrl_byp_o.wfi_stall = (id_ex_pipe_i.sys_en && id_ex_pipe_i.sys_wfi_insn && id_ex_pipe_i.instr_valid);
 
+  // Stall ID when mnxti CSR is accessed in EX
+  // This is needed because the data bypass from EX uses csr_rdata, and for mnxti this is actually mstatus and not the result
+  // that will be written to the register file. Could be optimized to only stall when the result from the CSR instruction is used in ID,
+  // but the common usecase is a CSR access followed by a branch using the mnxti result in the RF, so it would likely stall in most cases anyway.
+  assign ctrl_byp_o.mnxti_stall = (id_ex_pipe_i.csr_en && (id_ex_pipe_i.alu_operand_b[11:0] == CSR_MNXTI) && id_ex_pipe_i.instr_valid);
```
Silabs-ArjanB, line 136: Don't do an alu_operand_b comparison here, but instead add a signal similar to csr_counter_read_i as this address comparison is already performed in the CS registers module.

rtl/cv32e40x_controller_bypass.sv
```diff
@@ -175,7 +181,10 @@ module cv32e40x_controller_bypass import cv32e40x_pkg::*;
 
     // Stall because of jalr path. Stall if a result is to be forwarded to the PC except if result from WB is an ALU result.
     // No need to deassert anything in ID as ID stage is stalled anyway. alu_jmpr_id_i implies rf_re_id_i[0].
-    if (alu_jmpr_id_i && alu_en_raw_id_i && ((rf_we_wb && rf_rd_wb_jalr_match && lsu_en_wb) || (rf_we_ex && rf_rd_ex_jalr_match))) begin
+    if ((alu_jmpr_id_i && alu_en_raw_id_i) &&
+         ((rf_we_wb && rf_rd_wb_jalr_match && lsu_en_wb) ||
+          (rf_we_wb && rf_rd_wb_jalr_match && (ex_wb_pipe_i.csr_addr == CSR_MNXTI) && ex_wb_pipe_i.csr_en) ||
```
Silabs-ArjanB, line 186: Wouldn't it be cheaper if the CSR_MNXTI comparison from the CS register file is re-used here (delayed to have the correct timing)?

Why do we actually need this MNXTI based stall (as the required bypass is actually in the RTL)? Please add a comment explaning why.

bhv/cv32e40x_wrapper.sv
```diff
@@ -259,6 +259,13 @@ module cv32e40x_wrapper
                 .id_stage_multi_cycle_id_stall    (core_i.id_stage_i.multi_cycle_id_stall),
 
                 .id_stage_id_valid                (core_i.id_stage_i.id_valid_o),
+                .alu_op_a_mux_sel                 (core_i.id_stage_i.alu_op_a_mux_sel),
```
Silabs-ArjanB, line 262: All the signals that were added here should have _id_i postfix for their pin names

sva/cv32e40x_core_sva.sv
```diff
@@ -361,6 +374,35 @@ always_ff @(posedge clk , negedge rst_ni)
   endgenerate
 
 
+  // Check that operand_a data forwarded from EX to ID actually is written to RF in WB
```
Silabs-ArjanB, line 377: Can you add the equivalent assertions when forwarding from WB?

Silabs-ArjanB (changes requested):

rtl/cv32e40x_controller_bypass.sv
silabs-oysteink: Fixed

rtl/cv32e40x_controller_bypass.sv
silabs-oysteink: Added comment. The reason is that the jalr_fw is using the flopped ex_wb_pipe.rf_wdata (as opposed to the wb_stage.rf_wdata_o which causes timing paths from data_rvalid_i). If forwarding the mnxti read data the jalr_fw mux would need to also consider the pointer address from cs_register, and the use case for this CSR would likely not generate jumps following a CSR access to mnxti. 

I agree on sharing the logic, I will implement that.

bhv/cv32e40x_wrapper.sv
silabs-oysteink: Fixed

sva/cv32e40x_core_sva.sv
silabs-oysteink: Fixed

rtl/cv32e40x_controller_bypass.sv
Silabs-ArjanB: Thanks for the explanation. That is okay then; I had not seen that the registered rf_wdata was used in that bypass. This actually also means that there is a jalr_fw bug in case there is a result incoming via XIF in the WB stage that is supposed to be used by the JALR. Can you just add a todo for that in the code (no need to fix it as the XIF logic will get reorganized all together).

rtl/cv32e40x_controller_bypass.sv
```diff
@@ -181,9 +182,12 @@ module cv32e40x_controller_bypass import cv32e40x_pkg::*;
 
     // Stall because of jalr path. Stall if a result is to be forwarded to the PC except if result from WB is an ALU result.
     // No need to deassert anything in ID as ID stage is stalled anyway. alu_jmpr_id_i implies rf_re_id_i[0].
+    // Also stalling when a CSR access to MNXTI is in WB. This is because the jalr forward uses the ex_wb_pipe.rf_wdata as data,
+    // and forwarding the CSR result (pointer address for mnxti) would need the forward mux to also include the pointer address from cs_registers.
+    // Cannot use wb_stage.rf_wdata_o due to the timing path from the data OBI.
     if ((alu_jmpr_id_i && alu_en_raw_id_i) &&
          ((rf_we_wb && rf_rd_wb_jalr_match && lsu_en_wb) ||
-          (rf_we_wb && rf_rd_wb_jalr_match && (ex_wb_pipe_i.csr_addr == CSR_MNXTI) && ex_wb_pipe_i.csr_en) ||
+          (rf_we_wb && rf_rd_wb_jalr_match && ex_wb_pipe_i.csr_mnxti_access) ||
```
Silabs-ArjanB, line 190: Still needs qualification with ex_wb_pipe_i.csr_en

rtl/cv32e40x_controller_bypass.sv
silabs-oysteink: Todo added

Silabs-ArjanB (changes requested):

Silabs-ArjanB (approved):

PR #512, opened 2022-04-20
Removed nmi_addr_i; removed USE_DEPRECATED_FEATURE_SET parameter

Silabs-ArjanB:
SEC equivalent if USE_DEPRECATED_FEATURE_SET = 0.
Did not run assertion check flow as that is currently not compiling. 
Note that the a_rvfi_nmip_nmi_handler was fixed as well. It was not correct for either setting of USE_DEPRECATED_FEATURE_SET. I don't know why these assertions are excluded from formal runs (I will check with Halfdan).

### DIFF
--- a/bhv/cv32e40x_wrapper.sv
+++ b/bhv/cv32e40x_wrapper.sv
@@ -80,7 +80,6 @@ module cv32e40x_wrapper
   input  logic [31:0] mhartid_i,
   input  logic  [3:0] mimpid_patch_i,
   input  logic [31:0] dm_exception_addr_i,
-  input  logic [31:0] nmi_addr_i,
 
   // Instruction memory interface
   output logic        instr_req_o,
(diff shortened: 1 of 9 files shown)


bhv/cv32e40x_wrapper.sv
```diff
@@ -338,7 +337,7 @@ module cv32e40x_wrapper
       rvfi_sva(.irq_ack(core_i.irq_ack),
                .dbg_ack(core_i.dbg_ack),
                .ebreak_in_wb_i(core_i.controller_i.controller_fsm_i.ebreak_in_wb),
-               .nmi_addr_i(core_i.nmi_addr_i),
+               .mtvec_addr_i(core_i.mtvt_addr),
```
silabs-oysteink, line 340: This seems wrong, should it be connected to core_i.mtvec_addr?

silabs-oysteink (changes requested):

bhv/cv32e40x_wrapper.sv
Silabs-ArjanB: Fixed now

silabs-oysteink (approved):

PR #514, opened 2022-04-22
Bugfixes after running formal 

silabs-oysteink:
CLIC SHV  + single stepping should now work.
Disabled trigger matches for CLIC pointers (awaiting spec change for CLIC in this regards)
Various other small bugfixes and assertion updates.

### DIFF
--- a/sva/cv32e40x_controller_fsm_sva.sv
+++ b/sva/cv32e40x_controller_fsm_sva.sv
@@ -28,7 +28,9 @@
 module cv32e40x_controller_fsm_sva
   import uvm_pkg::*;
   import cv32e40x_pkg::*;
-  #(  parameter bit X_EXT     = 1'b0)
+  #(  parameter bit X_EXT     = 1'b0,
+      parameter bit SMCLIC    = 1'b0
+  )
 (
   input logic           clk,
   input logic           rst_n,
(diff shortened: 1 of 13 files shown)


sva/cv32e40x_controller_fsm_sva.sv
```diff
@@ -511,11 +513,25 @@ endgenerate
...
     else `uvm_error("controller", "LSU write buffer not empty when fetching CLIC pointer")
+
+  // After a pc_set to PC_TRAP_CLICV, only the following jump targets are allowed:
+  // PC_TRAP_CLICV_TGT: Normal execution, the pointer target is being fetched
+  // PC_TRAP_EXC:       The pointer fetch has a synchronous exception
+  // PC_TRAP_NMI:       The pointer fetch has a bus error. Todo: Change if CLIC spec stops using data access for pointer fetch.
+  a_clicv_next_pc_set:
+  assert property (@(posedge clk) disable iff (!rst_n)
+                  (ctrl_fsm_o.pc_set && (ctrl_fsm_o.pc_mux == PC_TRAP_CLICV))
+                  |=> !ctrl_fsm_o.pc_set until (ctrl_fsm_o.pc_set && ((ctrl_fsm_o.pc_mux == PC_TRAP_CLICV_TGT) ||
```
Silabs-ArjanB, line 530: It seems a little inconsistent: Here and in th CS register file single stepping debug entry is not allowed, but in the controller FSM it is actually implemented. Okay for now though as there is a todo on handling of single stepping for pointer fetches.

Silabs-ArjanB (approved):

PR #529, opened 2022-05-05
Corrected RW into WARL for mseccfg.RLB, mseccfg.MMWP, mseccfg.MML, pm…

Silabs-ArjanB:
…pxcfg.RWX, pmpaddrx.ADDRESS, mtvec.BASE


silabs-robin (approved):
I read the diff and double checked a few uncertainties. It looks alright to me.

PR #531, opened 2022-05-06
Implemented cm.lbu and cm.lhu from Zc C2.

silabs-oysteink:
+ Bugfix for c.sh, did not flag illegal instruction if bit 6 was set.

SEC clean when ZC_EXT=0

### DIFF
--- a/rtl/cv32e40x_compressed_decoder.sv
+++ b/rtl/cv32e40x_compressed_decoder.sv
@@ -119,6 +119,8 @@ module cv32e40x_compressed_decoder import cv32e40x_pkg::*;
                   3'b011: begin
                     // c.sh -> sh rs2', imm(rs1')
                     instr_o.bus_resp.rdata = {7'b0, 2'b01, instr[4:2], 2'b01, instr[9:7], 3'b001, 3'b000, instr[5], 1'b0, OPCODE_STORE};
+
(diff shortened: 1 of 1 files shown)


rtl/cv32e40x_compressed_decoder.sv
```diff
@@ -339,12 +341,34 @@ module cv32e40x_compressed_decoder import cv32e40x_pkg::*;
...
+            end
             3'b010: begin
               // c.lwsp -> lw rd, imm(x2)
               instr_o.bus_resp.rdata = {4'b0, instr[3:2], instr[12], instr[6:4], 2'b00, 5'h02, 3'b010, instr[11:7], OPCODE_LOAD};
               if (instr[11:7] == 5'b0)  illegal_instr_o = 1'b1;
             end
 
+            3'b011: begin
+              if (ZC_EXT) begin
+                // cm.lhu rd', uimm(rs1') -> lhu rd', uimm(rs1')
```
Silabs-ArjanB, line 377: Shouldn't this instruction be in the 3'b001 case?

Silabs-ArjanB (changes requested):

rtl/cv32e40x_compressed_decoder.sv
silabs-oysteink: Oh, you're right. I will fix it.

PR #541, opened 2022-05-13
Changed fetching of CLIC pointers to use instruction instead of data …

silabs-oysteink:
…fetches through the intruction interface.

Non SEC as this affects the prot bits and PMA results based on the prot bits.

### DIFF
--- a/rtl/cv32e40x_if_stage.sv
+++ b/rtl/cv32e40x_if_stage.sv
@@ -97,7 +97,6 @@ module cv32e40x_if_stage import cv32e40x_pkg::*;
   logic              prefetch_trans_valid;
   logic              prefetch_trans_ready;
   logic [31:0]       prefetch_trans_addr;
-  logic              prefetch_trans_data_access;
   inst_resp_t        prefetch_inst_resp;
   logic              prefetch_one_txn_pend_n;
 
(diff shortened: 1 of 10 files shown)


rtl/cv32e40x_if_stage.sv
```diff
@@ -281,14 +278,24 @@ module cv32e40x_if_stage import cv32e40x_pkg::*;
...
-        if_id_pipe_o.compressed_instr <= prefetch_instr.bus_resp.rdata[15:0];
+        if_id_pipe_o.compressed_instr <= prefetch_instr.bus_resp.rdata[15:0]; // todo: clock gate if not compressed.
         if_id_pipe_o.trigger_match    <= trigger_match_i;
         if_id_pipe_o.xif_id           <= xif_id;
+
+        if (prefetch_is_ptr) begin
+          // Update pointer value
+          if_id_pipe_o.ptr                <= instr_decompressed.bus_resp.rdata;
+
+          // Need to update bus error status and mpu status, but may omit the 32-bit instruction word
```
Silabs-ArjanB, line 292: Is it important that the instruction word is not updated here? If so, add a comment.

Silabs-ArjanB (commented):
There is some logic in the write buffer related to the write buffer and clic pointer fetches that should no longer be relevant. That should get removed as well (but can be done later).

Silabs-ArjanB (approved):

rtl/cv32e40x_if_stage.sv
silabs-oysteink: For CLIC it doesn't matter much, but for Zc* cm.jt/cm.jalt we will need to not update the instruction word to ensure the link part of cm.jalt is performed correctly in the second half.

PR #552, opened 2022-05-25
Zc* table jumps

silabs-oysteink:
Fully implemented table jumps for Zc*.

SEC clean when ZC_EXT=0


rtl/cv32e40x_if_stage.sv
```diff
@@ -114,27 +122,32 @@ module cv32e40x_if_stage import cv32e40x_pkg::*;
...
+      PC_TRAP_EXC:   branch_addr_n = {mtvec_addr_i, 7'h0};                                        // All the exceptions go to base address
+      PC_TRAP_IRQ:   branch_addr_n = {mtvec_addr_i, ctrl_fsm_i.mtvec_pc_mux, 2'b00};     // interrupts are vectored
+      PC_TRAP_DBD:   branch_addr_n = {dm_halt_addr_i[31:2], 2'b0};
+      PC_TRAP_DBE:   branch_addr_n = {dm_exception_addr_i[31:2], 2'b0};
+      PC_TRAP_NMI:   branch_addr_n = {mtvec_addr_i, NMI_MTVEC_INDEX, 2'b00};
+      PC_TRAP_CLICV: branch_addr_n = {mtvt_addr_i, ctrl_fsm_i.mtvt_pc_mux[SMCLIC_ID_WIDTH-1:0], 2'b00};
+      // CLIC and Zc* spec requires to clear bit 0. This clearing is done in the alignment buffer.
+      PC_POINTER :   branch_addr_n = if_id_pipe_o.ptr;
+      // JVT + (index << 2)
+      PC_TBLJUMP :   branch_addr_n = (jvt_i.base << 6) + (if_id_pipe_o.instr.bus_resp.rdata[19:12] << 2);
```
Silabs-ArjanB, line 150: We don't want a "+" here.

This needs to look like:

branch_addr_n = {jvt_addr_i + ctrl_fsm_i.jvt_pc_mux[7:0], 2'b00};

I will file an issue in the Zc TG to see if jvt.base can be declared WARL.

rtl/cv32e40x_core.sv
```diff
@@ -193,6 +193,9 @@ module cv32e40x_core import cv32e40x_pkg::*;
   logic [24:0] mtvec_addr;
   logic [1:0]  mtvec_mode;
 
+  // JVT
+  jvt_t        jvt;
```
Silabs-ArjanB, line 197: logic [21:0] jvt_addr;

rtl/cv32e40x_core.sv
```diff
@@ -404,6 +407,7 @@ module cv32e40x_core import cv32e40x_pkg::*;
     .mepc_i              ( mepc                     ), // Exception PC (restore upon return from exception/interrupt)
     .mtvec_addr_i        ( mtvec_addr               ), // Exception/interrupt address (MSBs only)
     .mtvt_addr_i         ( mtvt_addr                ), // CLIC vector base
+    .jvt_i               ( jvt                      ),
```
Silabs-ArjanB, line 410: .jvt_addr_i (jvt_addr),

rtl/cv32e40x_core.sv
```diff
@@ -696,6 +700,8 @@ module cv32e40x_core import cv32e40x_pkg::*;
     .mtvec_addr_i               ( mtvec_addr_i[31:0]     ),
     .csr_mtvec_init_i           ( csr_mtvec_init_if      ),
 
+    .jvt_o                      ( jvt                    ),
```
Silabs-ArjanB, line 703: .jvt_addr_o (jvt_addr),

rtl/cv32e40x_cs_registers.sv
```diff
@@ -542,7 +544,7 @@ module cv32e40x_cs_registers import cv32e40x_pkg::*;
   always_comb
   begin
 
-    jvt_n                    = '0;
+    jvt_n                    = {csr_wdata_int[31:6], 6'b000000};
```
Silabs-ArjanB, line 547: This is as documented, but it leads to an adder in IF where we don't want it. This needs to become:

jvt_n = {csr_wdata_int[31:10], 10'b0000000000};

I will update spec and user manual.

rtl/cv32e40x_cs_registers.sv
```diff
@@ -911,6 +913,8 @@ module cv32e40x_cs_registers import cv32e40x_pkg::*;
     .rd_error_o (jvt_rd_error)
   );
 
+  assign jvt_o = jvt_t'(jvt_q);
```
Silabs-ArjanB, line 916: Only output address part

rtl/cv32e40x_id_stage.sv
```diff
@@ -605,7 +611,7 @@ module cv32e40x_id_stage import cv32e40x_pkg::*;
 
         id_ex_pipe_o.illegal_insn           <= illegal_insn && !xif_insn_accept;
 
-        id_ex_pipe_o.rf_we                  <= rf_we;
+        id_ex_pipe_o.rf_we                  <= tbljmp_first ? 1'b0 : rf_we;  // Only the last part of table jumps write to the link register
```
Silabs-ArjanB, line 614: Can't this be done inside the decoder or does it not fit there?

rtl/cv32e40x_if_stage.sv
```diff
@@ -251,14 +265,20 @@ module cv32e40x_if_stage import cv32e40x_pkg::*;
...
+  // todo: Factor in last operation of sequenced Zc* (push/pop)
+  assign last_op = tbljmp ? 1'b0 : 1'b1;
 
   // Populate instruction meta data
   instr_meta_t instr_meta_n;
   always_comb begin
     instr_meta_n = '0;
-    instr_meta_n.compressed    = instr_compressed_int;
-    instr_meta_n.clic_ptr      = prefetch_is_ptr;
+    instr_meta_n.compressed    = if_id_pipe_o.instr_meta.compressed;
```
Silabs-ArjanB, line 279: This code looks weird. I did not try to understand it, but not all bits in instr_meta_n flow in the same way through the design, which is suspicious. If it cannot made to be more regular it at least deserves some comments. (Maybe this is a 'hidden FSM')

rtl/cv32e40x_if_stage.sv
```diff
@@ -282,13 +303,29 @@ module cv32e40x_if_stage import cv32e40x_pkg::*;
...
         if_id_pipe_o.instr_meta       <= instr_meta_n;
         if_id_pipe_o.illegal_c_insn   <= illegal_c_insn;
-        if_id_pipe_o.pc               <= pc_if_o;
-        if_id_pipe_o.compressed_instr <= prefetch_instr.bus_resp.rdata[15:0]; // todo: clock gate if not compressed.
+
+
         if_id_pipe_o.trigger_match    <= trigger_match_i;
         if_id_pipe_o.xif_id           <= xif_id;
-        if_id_pipe_o.last_op          <= 1'b1;
+        if_id_pipe_o.last_op          <= last_op;
```
Silabs-ArjanB, line 310: What will happen if there is a trigger match on the fetch from the table? It needs to be discarded.

rtl/cv32e40x_if_stage.sv
```diff
@@ -282,13 +303,29 @@ module cv32e40x_if_stage import cv32e40x_pkg::*;
...
+
+        // No PC update for tablejump pointer, PC of instruction itself is needed later.
+        // No update to the meta compressed, as this is used in calculating the link address.
+        //   Any pointer could change instr_compressed_int and cause a wrong link address.
+        // No update to tbljmp flag, we want flag to be high for both operations.
+        if (!prefetch_is_tbljmp_ptr) begin
```
Silabs-ArjanB, line 316: How about the CLIC pointer fetch; should that also prevent certain updates?

rtl/cv32e40x_if_stage.sv
```diff
@@ -314,12 +351,14 @@ module cv32e40x_if_stage import cv32e40x_pkg::*;
...
-    .instr_is_ptr_i     ( prefetch_is_ptr         ),
+    .instr_is_ptr_i     ( ptr_in_if_o             ),
     .instr_o            ( instr_decompressed      ),
     .is_compressed_o    ( instr_compressed_int    ),
-    .illegal_instr_o    ( illegal_c_insn          )
+    .illegal_instr_o    ( illegal_c_insn          ),
+    .tbljmp_o           ( tbljmp_raw              )
   );
 
+  assign tbljmp = (instr_decompressed.bus_resp.err || (instr_decompressed.mpu_status != MPU_OK)) ? 1'b0 : tbljmp_raw;
```
Silabs-ArjanB, line 361: Add comment on 'why' this is needed

Silabs-ArjanB (changes requested):

rtl/cv32e40x_compressed_decoder.sv
```diff
@@ -438,13 +440,37 @@ module cv32e40x_compressed_decoder import cv32e40x_pkg::*;
               end
             end
 
+            3'b101: begin
+              if (ZC_EXT) begin
+                // The cm.jt and cm.jalt have no equivalent 32-bit instructions.
+                // Mapping to JAL anyway, but an extra control bit is set to indicate that these
+                // are table jumps as opposed to regular JAL instruction.
+                if (instr[12:8] == 5'b00000) begin
```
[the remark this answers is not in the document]
silabs-oysteink: Indeed, that looks much better.

rtl/cv32e40x_if_stage.sv
silabs-oysteink: I guess you mean "branch_addr_n = {jvt_addr_i, ctrl_fsm_i.jvt_pc_mux[7:0], 2'b00};" ? This was what I wanted to implement, but following the spec didn't leave enough bits tied to zero to make the concatenation.

rtl/cv32e40x_controller_fsm.sv
```diff
@@ -385,15 +394,20 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
...
+  assign tbljmp_in_ex_wb = ((id_ex_pipe_i.instr_meta.tbljmp && id_ex_pipe_i.last_op) || (ex_wb_pipe_i.instr_meta.tbljmp && ex_wb_pipe_i.last_op));
+
   // Allow interrupts to be taken only if there is no data request in WB,
   // and no trans_valid has been clocked from EX to environment.
   // Offloaded instructions in WB also block, as they cannot be killed after commit_kill=0 (EX stage)
   // LSU instructions which were suppressed due to previous exceptions or trigger match
   // will be interruptable as they were convered to NOP in ID stage.
   // The cycle after fencei enters WB, the fencei handshake will be initiated. This must complete and the fencei instruction must retire before allowing interrupts.
   // TODO:OK:low May allow interuption of Zce to idempotent memories
+  // Once the first part of a table jump has finished in WB, we are not allowed to take interrupts before the last part finishes. This can be detected when the last
```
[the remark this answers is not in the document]
silabs-oysteink: Split load/stores are currently handled by the interruptible_o from the load_store_unit. for push/pop and multi moves I think it should be enough to add a single bit to the pipeline registers that tells that it is a multi operation instruction. I can't (at the moment) see any reason to be able to distinguish between them other than being able to block debug and interrupts.

rtl/cv32e40x_id_stage.sv
silabs-oysteink: At first I thought it could you deasser_we, but that would also deassert alu_en and the jump would not be taken. It can for sure be done in the decoder with something like 

`rf_we = (deassert_we_i || tbljmp_first) ? 1'b0 : rf_we`

silabs-oysteink: Moved to decoder

rtl/cv32e40x_if_stage.sv
silabs-oysteink: Added comments to explain. The idea is that the fields 'compressed' and 'tbljmp' needs to retain their value for table jumps, even for the second operation where a pointer is in the predecoder. The fields will be updated in 'the regular way' for all other cases than tablejump pointer. I see that this may be rewritten to be more clear though.

rtl/cv32e40x_if_stage.sv
silabs-oysteink: Currently there are no triggers on the table fetch. The PC field (used in cs_registers for trigger matching) is kept identical for both the instruction itself and the table fetch. Whenever a pointer is fetched, the IF stage output 'ptr_in_if_o' will prevent trigger matches. Any trigger matches on the PC during the second operation would already have been taken during the first operation and the table fetch would not occur. I guess the spec isn't yet including information about trigger on table fetches?

rtl/cv32e40x_if_stage.sv
silabs-oysteink: It's not needed for functional correctness, but we can possibly save some power by clock gating fields like 'instr_meta.compressed' as it doesn't make sense for CLIC pointers. The field 'instr.bus_resp.rdata' is already gated for both pointer types.

rtl/cv32e40x_core.sv
silabs-oysteink: Fixed

rtl/cv32e40x_cs_registers.sv
silabs-oysteink: Fixed

rtl/cv32e40x_controller_fsm.sv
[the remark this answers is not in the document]
Silabs-ArjanB: Let's introduce an 'id' into the pipeline in a following PR to generalize the interrupt/debug behavior.

rtl/cv32e40x_if_stage.sv
Silabs-ArjanB: Okay

rtl/cv32e40x_if_stage.sv
Silabs-ArjanB: 0.70.3 does not contain the info related to trigger match. It is present in 0.70.4 and in our spec. It seems you already implemented the required behavior though (I did not yet see the related part in cs_registers when I wrote my question)

rtl/cv32e40x_cs_registers.sv
Silabs-ArjanB: Zc spec needs to change as well. I assume that will happen as the current spec is awkward (and also not inline with for example mtvec which is a similar CSR)

rtl/cv32e40x_controller_fsm.sv
```diff
@@ -360,7 +367,9 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
   // LSU will not be interruptible if the outstanding counter != 0, or
   // a trans_valid has been clocked without ex_valid && wb_ready handshake.
   // The cycle after fencei enters WB, the fencei handshake will be initiated. This must complete and the fencei instruction must retire before allowing debug.
-  assign debug_allowed = lsu_interruptible_i && !fencei_ongoing && !xif_in_wb && !pointer_in_pipeline;
+  // Once the first part of a table jump has finished in WB, we are not allowed to take debug before the last part finishes. This can be detected when the last
+  // part of a table jump is in either EX or WB.
+  assign debug_allowed = lsu_interruptible_i && !fencei_ongoing && !xif_in_wb && !pointer_in_pipeline && !tbljmp_in_ex_wb;
```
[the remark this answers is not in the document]
Silabs-ArjanB: I think this code could be more generic (I put it on my own list of things to possibly look at; okay for now)

rtl/cv32e40x_decoder.sv
```diff
@@ -232,7 +235,10 @@ module cv32e40x_decoder import cv32e40x_pkg::*;
   assign lsu_en_o = deassert_we_i ? 1'b0 : lsu_en;
 
   assign csr_en_o = deassert_we_i ? 1'b0 : csr_en;
-  assign rf_we_o  = deassert_we_i ? 1'b0 : rf_we;
+
+  // rf_we is deasserted with deassert_we as all other enables
+  // but also for the first part of a table jump (only the last part write to the link register)
+  assign rf_we_o  = (deassert_we_i || tbljmp_first_i) ? 1'b0 : rf_we;
```
Silabs-ArjanB, line 241: I think this could go all the way into I decoder. Okay like this.

Silabs-ArjanB (approved):

PR #572, opened 2022-06-05
Remove double pkg import from compressed decoder

spersvold:
Signed-off-by: Steffen Persvold <spersvold@gmail.com>


Silabs-ArjanB (approved):

Silabs-ArjanB:
@spersvold Thank you very much for spotting this and fixing it as well!

PR #580, opened 2022-06-10
Update XIF interface

michael-platzer:
Hi,

This PR updates the XIF interface definition, the LSU, and the WB stage to the changes in openhwgroup/core-v-xif@e65cc88 and openhwgroup/core-v-xif@6c257cd.

XIF memory requests are no longer aligned by the CPU. Instead, data and byte enable signals (`wdata`, `rdata`, and `be`) are assumed to be already aligned to the width of the memory bus (any required alignment and splitting must be performed by the coprocessor).

The XIF memory request's transaction attribute `attr[1]` (indicates whether the request is naturally aligned) is included in `misaligned_access` and thus made available to the PMA.

**Question:** I am unsure how to handle the other transaction attribute (`attr[0]`, indicating whether the transaction was modified by the coprocessor) in CV32E40X. In the current PR this attribute is ignored. Does it need to be taken in account, and, if yes, what would be the consequence if it is set for a transaction?

### DIFF
--- a/rtl/cv32e40x_load_store_unit.sv
+++ b/rtl/cv32e40x_load_store_unit.sv
@@ -354,7 +354,8 @@ module cv32e40x_load_store_unit import cv32e40x_pkg::*;
   assign lsu_rdata_1_o = rdata_ext;
 
   // misaligned_access is high for both transfers of a misaligned transfer
-  assign misaligned_access = split_q || lsu_split_0_o || misaligned_halfword || (xif_req && xif_mem_if.mem_req.attr[1]);
+  // TODO: Give MPU a separate modified_access_i input
+  assign misaligned_access = split_q || lsu_split_0_o || misaligned_halfword || (xif_req && (xif_mem_if.mem_req.attr[0] || xif_mem_if.mem_req.attr[1]));
 
   // Check for misaligned accesses that need a second memory access
   // If one is detected, this is signaled with lsu_split_0_o.
--- a/rtl/cv32e40x_load_store_unit.sv
+++ b/rtl/cv32e40x_load_store_unit.sv
@@ -153,9 +153,13 @@ module cv32e40x_load_store_unit import cv32e40x_pkg::*;
   // Generate address from operands (atomic memory transactions do not use an address offset computation)
   always_comb begin
     if (xif_req) begin
-      trans.addr  = xif_mem_if.mem_req.addr;
+      // XIF memory data signals (wdata, rdata, & be) are already aligned to X_MEM_WIDTH; clear
+      // lower bits of address to avoid further modification by alignment logic.
+      trans.addr  = {xif_mem_if.mem_req.addr[31:2], 2'b00};
       trans.we    = xif_mem_if.mem_req.we;
-      trans.size  = 2'b10;                    // XIF memory requests always use maximum size
+      // Note: The `size` field in the XIF memory request is the size of individual elements; more
+      // bits may be set in `be`; set trans.size always to the bus width to allow any `be` pattern.
+      trans.size  = 2'b10;
       trans.wdata = xif_mem_if.mem_req.wdata;
       trans.mode  = xif_mem_if.mem_req.mode;  // TODO use mode from XIF request or force machine mode?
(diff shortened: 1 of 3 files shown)


rtl/cv32e40x_load_store_unit.sv
```diff
@@ -153,9 +153,13 @@ module cv32e40x_load_store_unit import cv32e40x_pkg::*;
...
     if (xif_req) begin
-      trans.addr  = xif_mem_if.mem_req.addr;
+      // XIF memory data signals (wdata, rdata, & be) are already aligned to X_MEM_WIDTH; clear
+      // lower bits of address to avoid further modification by alignment logic.
+      trans.addr  = {xif_mem_if.mem_req.addr[31:2], 2'b00};
       trans.we    = xif_mem_if.mem_req.we;
-      trans.size  = 2'b10;                    // XIF memory requests always use maximum size
+      // Note: The `size` field in the XIF memory request is the size of individual elements; more
+      // bits may be set in `be`; set trans.size always to the bus width to allow any `be` pattern.
+      trans.size  = 2'b10;
```
Silabs-ArjanB, line 162: Shouldn't this now become:

trans.size= xif_mem_if.mem_req.size; // todo: signal width mismatch will get fixed later

(because the LSU should not assume that it can freely 'modify' transactions from the coprocessor; it should basically simply pass on the coprocessor transactions without further modifications (assuming not blocked by the PMA).

rtl/cv32e40x_load_store_unit.sv
```diff
@@ -153,9 +153,13 @@ module cv32e40x_load_store_unit import cv32e40x_pkg::*;
   // Generate address from operands (atomic memory transactions do not use an address offset computation)
   always_comb begin
     if (xif_req) begin
-      trans.addr  = xif_mem_if.mem_req.addr;
+      // XIF memory data signals (wdata, rdata, & be) are already aligned to X_MEM_WIDTH; clear
+      // lower bits of address to avoid further modification by alignment logic.
+      trans.addr  = {xif_mem_if.mem_req.addr[31:2], 2'b00};
```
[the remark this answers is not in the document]
michael-platzer: I have now moved the multiplexers from the `trans.*` to the `align_trans.*` signals, as you suggested. In hindsight, this is a more sensible solution. I agree that bypassing the alignment logic of the core makes more sense now that the coprocessor aligns its own memory requests itself.

Regarding the read data zero/sign extension and alignment logic, it appears that the `lsu_size_q` and `rdata_offset_q` are used for that purpose only. The latest commit sets `lsu_size_q` to `2'b10` and `rdata_offset_q` to 0 for XIF requests, thus preventing any modification of the read data. Since these signals are not used for any other purpose there should be no unintended side effects. However, if you prefer I can modify the read data extension and alignment logic instead and set these signals to their actual values.

rtl/cv32e40x_load_store_unit.sv
michael-platzer: After moving the multiplexers to the `align_trans.*` signals, and since these signals do not have a `size` field, assigning the `size` value of the XIF request appears to be no longer required. However, please see my other comment regarding the `lsu_size_q` value and let me know whether you would prefer to assign the XIF size field to `lsu_size_q` instead.

michael-platzer:
Hi @Silabs-ArjanB,

thanks for your feedback! Please see my answers to your comments above regarding bypassing the alignment logic.

I have also updated the assignment of the `misaligned_access` signal as you suggested.

rtl/cv32e40x_load_store_unit.sv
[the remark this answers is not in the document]
Silabs-ArjanB: The treatment of the extension logic is okay for now.

rtl/cv32e40x_load_store_unit.sv
Silabs-ArjanB: I agree that XIF size is not used/needed at the moment. In the future we might expand the OBI interface to present both 'size' and 'be'.

Silabs-ArjanB (approved):

PR #607, opened 2022-07-05
Sequencer integration

silabs-oysteink:
Integrating sequencer with pipeline.
- Not allowing interrupts to break sequences
- Not allowing NMI to break sequences
- Not allowing debug to break sequences
- Exception will be taken on any operation from WB
- Halting of ID stage not allowed if ID stage is in the middle of a sequence (to avoid deadlocks where the core is waiting for a sequence to finish before it can take an interrupt)
- Updated logic to halt IF due to single step to account for multiple operations from the sequencer
-
SEC clean when localparam ZC_EXT = 0 and correct OBI behavior is assumed (no instr_rvalid_i before instr_req_o)

### DIFF
--- a/rtl/cv32e40x_controller_fsm.sv
+++ b/rtl/cv32e40x_controller_fsm.sv
@@ -47,7 +47,9 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
   input  ctrl_byp_t   ctrl_byp_i,
 
   // From IF stage
-  input logic [31:0]   pc_if_i,
+  input  logic [31:0] pc_if_i,
+  input  logic        first_op_if_i,              // IF stage is handling the first operation of a sequence.
+  input  logic        last_op_if_i,               // IF stage is handling the last operation of a sequence.
 
   // From ID stage
   input  if_id_pipe_t if_id_pipe_i,
@@ -55,11 +57,13 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
   input  logic        sys_mret_id_i,              // mret in ID
   input  logic        alu_en_id_i,                // alu_en qualifier for jumps
   input  logic        sys_en_id_i,                // sys_en qualifier for mret
+  input  logic        first_op_id_i,              // ID stage is handling the first operation of a sequence
 
   // From EX stage
   input  id_ex_pipe_t id_ex_pipe_i,
   input  logic        branch_decision_ex_i,       // branch decision signal from EX ALU
   input  logic        last_op_ex_i,               // EX stage contains the last operation of an instruction
+  input  logic        first_op_ex_i,              // EX stage is handling the first operation of a sequence
 
   // From WB stage
   input  ex_wb_pipe_t ex_wb_pipe_i,
@@ -209,12 +213,16 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
   // Do not count if halted or killed
   logic       wb_counter_event_gated;
 
-  // Detect uninterruptible table jumps
-  logic       tbljmp_in_ex_wb;
-
   // Flop for acking flush requests due to CSR writes
   logic       csr_flush_ack_n;
   logic       csr_flush_ack_q;
+
+  // Flag for checking if pipeline can be killed due to multi op instructions
+  logic       sequence_interruptible;
+
+  // Flg for checking if ID stage can be halted
+  // Used to not halt sequences in the middle, potentially causing deadlocks
+  logic       id_stage_haltable;
 
   assign fencei_ready = !lsu_busy_i;
 
@@ -271,6 +279,8 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
   assign branch_taken_ex = branch_in_ex && !branch_taken_q;
 
   // Exception in WB if the following evaluates to 1
+  // Not checking for ex_wb_pipe_i.last_op to enable exceptions to be taken as soon as possible for
+  // split load/stores or Zc sequences.
   assign exception_in_wb = ((ex_wb_pipe_i.instr.mpu_status != MPU_OK)                              ||
                              ex_wb_pipe_i.instr.bus_resp.err                                       ||
                             ex_wb_pipe_i.illegal_insn                                              ||
@@ -386,7 +396,7 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
   // The cycle after fencei enters WB, the fencei handshake will be initiated. This must complete and the fencei instruction must retire before allowing debug.
   // Once the first part of a table jump has finished in WB, we are not allowed to take debug before the last part finishes. This can be detected when the last
   // part of a table jump is in either EX or WB.
-  assign debug_allowed = lsu_interruptible_i && !fencei_ongoing && !xif_in_wb && !pointer_in_pipeline && !tbljmp_in_ex_wb;
+  assign debug_allowed = lsu_interruptible_i && !fencei_ongoing && !xif_in_wb && !pointer_in_pipeline && sequence_interruptible;
 
   // Debug pending for any other reason than single step
   assign pending_debug = (trigger_match_in_wb) ||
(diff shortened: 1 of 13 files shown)


rtl/cv32e40x_controller_fsm.sv
```diff
@@ -499,9 +537,12 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
     // Interrupts: Machine mode: Prevent issuing new instructions until we get an interruptible bubble.
     //             Debug mode:   Interrupts are not allowed during debug. Cannot halt ID stage in such a case
     //                           since the dret that brings the core out of debug mode may never get passed a halted ID stage.
+    //             Sequences:    If we need to halt for debug or interrupt not allowed due to a sequence, we must check if we can
```
Silabs-ArjanB, line 540: passed -> past

rtl/cv32e40x_if_stage.sv
```diff
@@ -300,12 +295,17 @@ module cv32e40x_if_stage import cv32e40x_pkg::*;
 
   // Last operation of table jumps are set when the pointer is fed to ID stage
   // tbljmp is set when a cm.jt or cm.jalt is decoded in the compressed decoder.
-  // todo: Factor in last operation of sequenced Zc* (push/pop)
-  assign last_op = tbljmp ? 1'b0 : 1'b1;
+  // Trigger matches will get all write enables deasserted, and debug entered before executing
+  // any operations. A trigger match is a last_op by definition.
+  // todo: Factor CLIC pointers?
+  assign last_op_o = trigger_match_i ? 1'b1 :
```
Silabs-ArjanB, line 301: What is supposed to happen for a trigger match on a sequenced instruction? Will the sequencer keep on sequencing?

rtl/cv32e40x_if_stage.sv
```diff
@@ -300,12 +295,17 @@ module cv32e40x_if_stage import cv32e40x_pkg::*;
...
+                     tbljmp ? 1'b0 :
+                     seq_valid ? seq_last : 1'b1;
 
   // Flag first operation of a sequence.
-  // todo: factor in seq_first. For now only tablejump pointer may be the only non-first operation.
-  assign first_op = prefetch_is_tbljmp_ptr ? 1'b0 : 1'b1;
+  // todo: factor in CLIC pointers?
+  assign first_op_o = prefetch_is_tbljmp_ptr ? 1'b0 : seq_first;
```
Silabs-ArjanB, line 310: first and last do not seem to get handled consistently. For example seq_valid is used to decide whether to use seq_last or not, whereas seq_first seems always used.

In general the IF stage can have various reasons/submodules to drive first_op/last_op and I would expect these modules to be clear and to be handled similarly for first_op and last_op.

rtl/cv32e40x_if_stage.sv
```diff
@@ -417,26 +417,29 @@ module cv32e40x_if_stage import cv32e40x_pkg::*;
...
+        .seq_first_o        ( seq_first               ),
+        .seq_last_o         ( seq_last                )
       );
     end else begin : gen_no_seq
       assign seq_valid = 1'b0;
       assign seq_last  = 1'b0;
       assign seq_instr = '0;
       assign seq_ready = 1'b1;
-      assign seq_first = 1'b0;
+      assign seq_first = 1'b1; // Tie high to enable default first_op when ZC_EXT=0
```
Silabs-ArjanB, line 440: Looks a bit like a hack because of a non-consistent strategy with respect to merging first/last signals of various modules.

Silabs-ArjanB (changes requested):

rtl/cv32e40x_controller_fsm.sv
```diff
@@ -456,6 +463,37 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
...
+  // Used to determine if we are allowed to take debug or interrupts
+  always_comb begin
+    if (ex_wb_pipe_i.instr_valid) begin
+      sequence_interruptible = ex_wb_pipe_i.first_op;
+    end else if (id_ex_pipe_i.instr_valid) begin
+      sequence_interruptible = first_op_ex_i;
+    end else if (if_id_pipe_i.instr_valid) begin
+      sequence_interruptible = first_op_id_i;
+    end else begin
+      sequence_interruptible = first_op_if_i;
```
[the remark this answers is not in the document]
silabs-oysteink: If there is a sequenced instruction IF, then the first_op_if will be driven to match the current state of the sequencer. If there is no instruction in IF, the sequencer will by default drive first_op to 1. As soon as a sequenced instruction enter IF, the stage will contain that instruction until it is done. As such, if the IF stage does not contain any instruction, first_op will be 1 and a sequence is not in IF at all.

rtl/cv32e40x_if_stage.sv
silabs-oysteink: The sequencer will keep on sequencing, but all write enables will be deasserted in the ID stage. As soon as the first operation reaches WB, a debug entry will be performed with the PC of the sequenced instruction in DPC.

rtl/cv32e40x_controller_fsm.sv
```diff
@@ -209,13 +213,17 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
...
-  logic       tbljmp_in_ex_wb;
-
   // Flop for acking flush requests due to CSR writes
   logic       csr_flush_ack_n;
   logic       csr_flush_ack_q;
 
+  // Flag for checking if pipeline can be killed due to multi op instructions
+  logic       sequence_interruptible;
+
+  // Flg for checking if ID stage can be halted
```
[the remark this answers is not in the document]
silabs-oysteink: Fixed

rtl/cv32e40x_controller_fsm.sv
```diff
@@ -209,13 +213,17 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
...
   logic       wb_counter_event_gated;
 
-  // Detect uninterruptible table jumps
-  logic       tbljmp_in_ex_wb;
-
   // Flop for acking flush requests due to CSR writes
   logic       csr_flush_ack_n;
   logic       csr_flush_ack_q;
 
+  // Flag for checking if pipeline can be killed due to multi op instructions
```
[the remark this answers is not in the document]
silabs-oysteink: Fixed

rtl/cv32e40x_controller_fsm.sv
```diff
@@ -456,6 +463,37 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
...
+    end
+  end
+
+  // Check if we are allowed to halt the ID stage.
+  // If ID stage contains a non-first operation, we cannot halt ID as that
+  // could cause a deadlock because a sequence cannot finish through ID.
+  // If ID stage does not contain a valid instruction, the same check is performed
+  // for the IF stage (although this is likely not needed).
+  always_comb begin
+    id_stage_haltable = 1'b0;
```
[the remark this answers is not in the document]
silabs-oysteink: Removed

rtl/cv32e40x_if_stage.sv
silabs-oysteink: Agreed, they are not consistent. Making them consistent (including using seq_valid for calculating first_op) seems very difficult without introducing combinatorial loops.

rtl/cv32e40x_controller_fsm.sv
```diff
@@ -456,6 +463,37 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
...
+    if (if_id_pipe_i.instr_valid) begin
+      id_stage_haltable = first_op_id_i;
+    end else begin
+      // IF stage first_op defaults to 1'b1 if the stage does not hold a valid instruction.
```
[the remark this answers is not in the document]
silabs-oysteink: This comment was not very clear. I updated the comment now - "valid instruction" in this comment really meant "nothing is valid in IF". For table jump pointer, the IF stage is valid and will set first_op to 0.

PR #617, opened 2022-07-12
fix default assignments

davideschiavone:


rtl/cv32e40x_cs_registers.sv
```diff
@@ -1015,6 +1015,7 @@ module cv32e40x_cs_registers import cv32e40x_pkg::*;
           csr_wdata_int = csr_wdata;
           csr_we_int    = 1'b0;
         end
+        default:;
```
Silabs-ArjanB, line 1018: This case statement is already complete. I believe we would get a Lint warning when adding a default here.

rtl/cv32e40x_id_stage.sv
```diff
@@ -307,6 +307,7 @@ module cv32e40x_id_stage import cv32e40x_pkg::*;
     unique case (imm_a_mux_sel)
       IMMA_Z:      imm_a = imm_z_type;
       IMMA_ZERO:   imm_a = '0;
+      default:     imm_a = '0;
```
Silabs-ArjanB, line 310: This case statement is already complete. I believe we would get a Lint warning when adding a default here.

Silabs-ArjanB: However, if the above is true, then we should have a warning for jalr_fw_mux.

Silabs-ArjanB (commented):
We will Lint check first before we can merge this

Silabs-ArjanB:
@silabs-oysteink Can you check with our Lint setup if this PR would increase the number of default related warnings? Some of the fixes in this PR are definitely good, but maybe not all.

davideschiavone:
Hi @Silabs-ArjanB , indeed I got most of the time just warnings, even when the case was somehow complete (with Verilator)

Silabs-ArjanB:
@davideschiavone My point is that I believe that we sometimes get warnings because of adding a default (but we will recheck that).

davideschiavone:
ah no I don't think so (for my case I mean), because using the branch of the PR is shutting down those warnings, again, with Verilator. Other tools may behave differenlty

silabs-oysteink:
I think the changes are good. They do lead to an increased amount of lint warnings though, both for the fully covered cases with added default, and for those that are not full which now gets a 'reachable default'. I think it is better to be consistent with defaults, and instead waive warnings in lint if wanted.

PR #628, opened 2022-07-26
Redefined NMI target address for basic non-vectored mode and CLIC mode

Silabs-ArjanB:

### DIFF
--- a/docs/user_manual/source/control_status_registers.rst
+++ b/docs/user_manual/source/control_status_registers.rst
@@ -667,10 +667,14 @@ Detailed:
 The initial value of ``mtvec`` is equal to {**mtvec_addr_i[31:7]**, 5'b0, 2'b01}.
 
 When an exception or an interrupt is encountered, the core jumps to the corresponding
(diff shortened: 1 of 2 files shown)


silabs-hfegran (commented):
While I do not have any objections to the suggestion in the current form, I wonder if there is 
any way we can avoid trampoline code to get to the nmi handler? In the this suggested change, all cases will require some sort of  nmi -> offset(mtvec) = j nmi_handler -> nmi_handler sequence. This seems to be a necessary evil of tying nmi handlers to mtvec. 

I guess my concern is if this is a good idea? Are there any possible security implications by letting a potential glitch actually bypass the jump to an nmi handler, instead of actually forcing execution at the actual handler location?

Silabs-ArjanB:
Hi @silabs-hfegran , the reason to use mtvec is because it is cheaper and it allows moving of all exception and interrupt related locations at once. The NMIs important for security raise an alert on major alert as well and can therefore fully be handled in hardware without the need to rely on the jump chain that you describe.

PR #645, opened 2022-08-23
Moved decoding of tablejumps to the sequencer

silabs-oysteink:
Tablejump pointers are handled within the sequencer, but bypass the internal FSM of the sequencer.

SEC clean for both values of ZC_EXT.

### DIFF
--- a/rtl/cv32e40x_if_stage.sv
+++ b/rtl/cv32e40x_if_stage.sv
@@ -96,10 +96,6 @@ module cv32e40x_if_stage import cv32e40x_pkg::*;
   logic              prefetch_is_clic_ptr;
   logic              prefetch_is_tbljmp_ptr;
 
-  // flag for predecoded cm.jt / cm.jalt.
-  // Maps to custom use of JAL instruction
-  logic              tbljmp;
-
   logic              illegal_c_insn;
 
   inst_resp_t        instr_decompressed;
@@ -136,6 +132,7 @@ module cv32e40x_if_stage import cv32e40x_pkg::*;
   logic              seq_first;       // sequencer is outputting the first operation
   logic              seq_last;        // sequencer is outputting the last operation
   inst_resp_t        seq_instr;       // Instruction for sequenced operation
+  logic              seq_tbljmp;      // Sequenced instruction is a table jump
 
   // Fetch address selection
   always_comb
@@ -364,7 +361,7 @@ module cv32e40x_if_stage import cv32e40x_pkg::*;
           // Sequenced instructions are marked as illegal by the compressed decoder, however, the instr_compressed
           // flag is still set and can be used when propagating to ID.
           if_id_pipe_o.instr_meta.compressed <= instr_compressed;
-          if_id_pipe_o.instr_meta.tbljmp     <= tbljmp;
+          if_id_pipe_o.instr_meta.tbljmp     <= seq_tbljmp;
 
           // Only update compressed_instr for compressed instructions
           if (instr_compressed) begin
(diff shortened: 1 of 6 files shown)


rtl/cv32e40x_if_stage.sv
```diff
@@ -431,7 +423,8 @@ module cv32e40x_if_stage import cv32e40x_pkg::*;
         .rst_n              ( rst_n                   ),
 
         .instr_i            ( prefetch_instr          ),
-        .instr_is_ptr_i     ( ptr_in_if_o             ),
+        .instr_is_clic_ptr_i( prefetch_is_clic_ptr    ),
+        .instr_is_tblj_ptr_i( prefetch_is_tbljmp_ptr  ),
```
Silabs-ArjanB, line 427: instr_is_tblj_ptr_i -> instr_is_tbljmp_ptr_i (prevent changing names in hook ups as much as possible)

rtl/cv32e40x_if_stage.sv
```diff
@@ -443,14 +436,16 @@ module cv32e40x_if_stage import cv32e40x_pkg::*;
         .valid_o            ( seq_valid               ),
         .ready_o            ( seq_ready               ),
         .seq_first_o        ( seq_first               ),
-        .seq_last_o         ( seq_last                )
+        .seq_last_o         ( seq_last                ),
+        .seq_tbljmp_o       ( tbljmp                  )
```
Silabs-ArjanB, line 440: tbljmp -> seq_tbljmp (for consistency with other sequencer outputs) (not needed if this signal will again disappear or need to be renamed in the later change of moving some decoding out of the sequencer into the predecoder again)

rtl/cv32e40x_sequencer.sv
```diff
@@ -63,22 +65,38 @@ import cv32e40x_pkg::*;
...
   // Count number of instructions emitted and set next state for FSM.
   always_ff @(posedge clk, negedge rst_n) begin
     if (!rst_n) begin
       instr_cnt_q <= '0;
       seq_state_q <= S_IDLE;
     end else begin
       if (valid_o && ready_i) begin
-        if (!seq_last_o) begin
+        // Exclude tablejumps and tablejump pointers from increasing the counter.
+        // To remain SEC clean, the prefetcher is ack'ed on a tablejump. If the tablejump
```
Silabs-ArjanB, line 95: What if you would do a SEC run taking the data_req_o signals into account? Then it should be possible to be SEC clean without using && !seq_tbljmp_o && !instr_is_tblj_ptr_i. If not, then this code might just be masking a real issue (unless the difference is due to different register source usage and related stall cycles).

rtl/cv32e40x_sequencer.sv
```diff
@@ -255,7 +285,18 @@ import cv32e40x_pkg::*;
...
+          if (instr[9:8] == 2'b00) begin
+            // cm.jt -> JAL x0, index
+            instr_o.bus_resp.rdata = {13'b0000000000000, instr[7:2], 5'b00000, OPCODE_JAL};
+          end else begin
+            // cm.jalt -> JAL, x1, index
+            instr_o.bus_resp.rdata = {11'b00000000000, instr[9:2], 5'b00001, OPCODE_JAL};
+          end
+          // The second half of tablejumps (pointer) will not use the FSM (the jump will kill the sequencer anyway).
+          // Signalling ready here will acknowledge the prefetcher (to remain SEC clean).
+          ready_fsm = ready_i && !halt_i;
```
Silabs-ArjanB, line 298: This I want to understand better. It is not a goal in itself to remain SEC clean. What are the alternatives / tradeoffs here?

Silabs-ArjanB: Also the comment about SEC clean will lose all meaning after this PR has been merged

rtl/cv32e40x_sequencer.sv
```diff
@@ -328,43 +369,50 @@ import cv32e40x_pkg::*;
...
-      ready_o = 1'b1;
+      ready_fsm = 1'b1;
       seq_state_n = S_IDLE;
     end
   end
 
+  // Bypass the sequencer FSM when there is an incoming tablejump pointer.
+  // Tablejump pointers are the last/non-first of a sequence.
+  // While waiting for a tablejump pointer, we still need to obey halt/kill inputs for the ready_o.
+  assign seq_last_o  = instr_is_tblj_ptr_i ? 1'b1               : seq_last_fsm;
```
Silabs-ArjanB, line 413: Only 1 'space' before ':' (same for next line)

Silabs-ArjanB (changes requested):

rtl/cv32e40x_if_stage.sv
silabs-oysteink: If we move all decoding from the sequencer to the compressed_decoder (very likely), then this signal may come from any of those two modules. Likely most logical to let it come from the sequencer, I'll rename it.

rtl/cv32e40x_sequencer.sv
silabs-oysteink: The difference between setting ready or not at this point is the same as ack'ing or not ack'ing the prefetcher. This will make the instruction that comes directly after the table jump either be a different instruction (ack) or the tablejump itself (not acking). For regular execution, this doesn't matter as the instruction will be killed when the tablejump is in ID. If the tablejump gets a bus error or MPU error in the fetch stage, the tablejump will not kill IF while in ID, and the following instruction may cause observable side effects (instruction OBI due to jumps) before the pipeline is killed from WB.

Silabs-ArjanB: What would your preferred RTL look like here? Performance is not a consideration during bus errors. If you want the code changed later please just add a todo so that we can leave the initial PR sec clean

silabs-oysteink: I think it is makes sense to acknowledge the prefetcer, i.e keeping the code as is.

rtl/cv32e40x_sequencer.sv
silabs-oysteink: I started a SEC run with this exclusions removed and only checking data OBI outputs when they are used (taking req and we into account). It hasn't finished yet however. I propose to remove these exclusions later, when the SEC run has been confirmed.

Issue #589, opened 2022-06-20
mscratchcswl and mscratchcsw should operate on mscratch


Silabs-ArjanB:
mscratchcswl and mscratchcsw  are currently implemented as regular CSRs and not according to the pseudo code stated in the CLIC specification.

Take https://github.com/riscv/riscv-fast-interrupt/pull/241 into account as well.

Silabs-ArjanB:
@silabs-oysteink Has this been fixed now including the mentioned issue 241?

silabs-oysteink:
Code is actually using mcause.mpp, so we are not compliant with the issue mentioned.

Silabs-ArjanB:
@silabs-oysteink Should be a SEC clean change to get it fixed then; I will add it in the coming sprint.

silabs-oysteink:
This issue should now be fixed.

PR #661, opened 2022-09-07
Fix for issue #589.

silabs-oysteink:
Removed instances for mscratchcsw and mscratchcswl CSRs. These CSRs now conditionally operate on mscratch according to the CLIC spec.

### DIFF
--- a/rtl/cv32e40x_cs_registers.sv
+++ b/rtl/cv32e40x_cs_registers.sv
@@ -668,10 +668,10 @@ module cv32e40x_cs_registers import cv32e40x_pkg::*;
       mintthresh_n             = csr_wdata_int & CSR_MINTTHRESH_MASK;
       mintthresh_we            = 1'b0;
 
-      mscratchcsw_n            = mstatus_n; // mscratchcsw operates conditionally on mscratch
+      mscratchcsw_n            = mscratch_n; // mscratchcsw operates conditionally on mscratch
       mscratchcsw_we           = 1'b0;
 
-      mscratchcswl_n           = mstatus_n; // mscratchcswl operates conditionally on mscratch
+      mscratchcswl_n           = mscratch_n; // mscratchcswl operates conditionally on mscratch
       mscratchcswl_we          = 1'b0;
 
       mie_n                    = '0;
--- a/rtl/cv32e40x_cs_registers.sv
+++ b/rtl/cv32e40x_cs_registers.sv
@@ -202,10 +202,10 @@ module cv32e40x_cs_registers import cv32e40x_pkg::*;
   logic [31:0]                  mintthresh_q, mintthresh_n, mintthresh_rdata;
   logic                         mintthresh_we;
 
-  logic [31:0]                  mscratchcsw_q, mscratchcsw_n, mscratchcsw_rdata;
+  logic [31:0]                  mscratchcsw_n, mscratchcsw_rdata;
   logic                         mscratchcsw_we;
 
-  logic [31:0]                  mscratchcswl_q, mscratchcswl_n, mscratchcswl_rdata;
+  logic [31:0]                  mscratchcswl_n, mscratchcswl_rdata;
   logic                         mscratchcswl_we;
 
   logic [31:0]                  mclicbase_q, mclicbase_n, mclicbase_rdata;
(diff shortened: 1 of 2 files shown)


sva/cv32e40x_cs_registers_sva.sv
```diff
@@ -90,6 +96,38 @@ module cv32e40x_cs_registers_sva
...
+      else `uvm_error("cs_registers", "Mscratch not stable after mscratwchsw with mpp=M");
+
+    // Check that mscratch do not update due to mscratchcswl if the conditions are not right
+    property p_mscratchcswl_mscratch;
+      @(posedge clk) disable iff (!rst_n)
+      (  ex_wb_pipe_i.csr_en && (csr_waddr == CSR_MSCRATCHCSWL) && !((mcause_rdata.mpil == '0) != (mintstatus_rdata.mil == '0))
+        |=> $stable(mscratch_rdata));
+    endproperty;
+
+    a_mscratchcswl_mscratch: assert property(p_mscratchcswl_mscratch)
```
silabs-oivind, line 117: Should we have a similar assertion for mscratchsw ?

rtl/cv32e40x_cs_registers.sv
```diff
@@ -646,10 +668,10 @@ module cv32e40x_cs_registers import cv32e40x_pkg::*;
       mintthresh_n             = csr_wdata_int & CSR_MINTTHRESH_MASK;
       mintthresh_we            = 1'b0;
 
-      mscratchcsw_n            = csr_wdata_int; // todo: isssue 589
+      mscratchcsw_n            = mstatus_n; // mscratchcsw operates conditionally on mscratch
```
silabs-oivind, line 671: Should this be mscratch_n ? IF not, the comment is confusing.

rtl/cv32e40x_cs_registers.sv
```diff
@@ -646,10 +668,10 @@ module cv32e40x_cs_registers import cv32e40x_pkg::*;
...
       mscratchcsw_we           = 1'b0;
 
-      mscratchcswl_n           = csr_wdata_int; // todo: isssue 589
+      mscratchcswl_n           = mstatus_n; // mscratchcswl operates conditionally on mscratch
```
silabs-oivind, line 674: Same as above

sva/cv32e40x_cs_registers_sva.sv
silabs-oysteink: I will add that to the E40S, on E40X such an assertions would not be reachable since mcause.mpp can only be of value PRIV_LVL_M.

rtl/cv32e40x_cs_registers.sv
silabs-oysteink: Good catch, not the first time I type the wrong CSR name. Proves why we need reviews. I'll push a fix

rtl/cv32e40x_cs_registers.sv
silabs-oysteink: Same answer as above

silabs-oivind (approved):

silabs-hfegran (commented):
From what I understand, and from your explanations, this looks good to me

rtl/cv32e40x_cs_registers.sv
```diff
@@ -1356,6 +1353,15 @@ module cv32e40x_cs_registers import cv32e40x_pkg::*;
...
   assign mnxti_rdata        = mnxti_irq_pending_i ? {mtvt_addr_o, mnxti_irq_id_i, 2'b00} : 32'h00000000;
 
+  // mscratchcsw_rdata breaks the regular convension for CSrs. Read data depend on mcause.mpp
+  // mscratch_rdata is returned if mcause.mpp differs from PRIV_LVL_M, otherwise rs1 is returned.
+  // This signal is only used by RVFI, and has WB timing (rs1 comes from ex_wb_pipe_i.csr_wdata, flopped version of id_ex_pipe.alu_operand_a)
+  assign mscratchcsw_rdata  = (mcause_rdata.mpp != PRIV_LVL_M) ? mscratch_rdata : ex_wb_pipe_i.csr_wdata;
+
+  // mscratchcswl_rdata breaks the regular convension for CSrs. Read data depend on mcause.pil and mintstatus.mil.
+  // This signal is only used by RVFI, and has WB timing (rs1 comes from ex_wb_pipe_i.csr_wdata, flopped version of id_ex_pipe.alu_operand_a)
+  assign mscratchcswl_rdata = ((mcause_rdata.mpil == '0) != (mintstatus_rdata.mil == 0)) ? mscratch_rdata : ex_wb_pipe_i.csr_wdata;
```
Silabs-ArjanB, line 1363: Not accurate because of the csr_wdata usage

sva/cv32e40x_cs_registers_sva.sv
```diff
@@ -46,7 +46,13 @@ module cv32e40x_cs_registers_sva
    input logic        clic_pa_valid_o,
    input mintstatus_t mintstatus_rdata,
    input privlvl_t    priv_lvl_n,
-   input privlvl_t    priv_lvl_rdata
+   input privlvl_t    priv_lvl_rdata,
+   input logic [31:0] mscratch_rdata,
+   input mcause_t     mcause_rdata,
```
Silabs-ArjanB, line 51: use mstatus instead of mcause everywhere related to mscratchcsw

sva/cv32e40x_cs_registers_sva.sv
```diff
@@ -90,6 +96,38 @@ module cv32e40x_cs_registers_sva
 
     a_htrap_interrupt_level: assert property(p_htrap_interrupt_level)
       else `uvm_error("cs_registers", "Horizontal trap taken caused interrupt level to change");
+
+    // Check that mscratch do not update due to mscratchcsw if the conditions are not right
+    property p_mscratchcsw_mscratch;
```
Silabs-ArjanB, line 101: These assertions are self fullfilling as they basically just match the RTL. Could an alternative implementation that is closer to the actual CLIC specification be used here instead? I.e. use the for this one use a mstatus bypass instead of the mcause/mstatus rdata which has the dangerous timing.

Silabs-ArjanB: Same remark for all assertions; use the 'bypass' (i.e. _q value of the CSR or to be written value of the CSR)

PR #664, opened 2022-09-08
Fix issue related to CLIC pointer fetch while MRET in WB

silabs-oivind:

### DIFF
--- a/rtl/cv32e40x_controller_fsm.sv
+++ b/rtl/cv32e40x_controller_fsm.sv
@@ -956,6 +956,12 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
           // The FSM must be in the FUNCTIONAL state to take the exception or NMI.
           // A faulted pointer (in ID) should not cause debug entry either,
         end
+
+        // Mret in WB restores CSR regs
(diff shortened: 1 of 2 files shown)


Silabs-ArjanB:
@silabs-oysteink, @silabs-oivind  Where can I find a description of the problem fixed with this PR?

silabs-oivind:
> @silabs-oysteink, @silabs-oivind Where can I find a description of the problem fixed with this PR?
@Silabs-ArjanB 
The problem here was that an mret in WB would not result in csr_restore_mret being asserted if the FSM was in the PONTER_FETCH state.
This in turn would result in mstatus, mintstatus and privilege level (on e40s) not being updated correctly upon mret.

Silabs-ArjanB:
The whole POINTER_FETCH state has become a bit convoluted now that mret is also being handled there. We really should have an assertion now that none of the other synchronous events that we check for in the FUNCTIONAL state can possibly be true while in the POINTER_FETCH state (to be sure that we are not missing handling of other things while waiting in POINTER_FETCH)

For example !(exception_in_wb && exception_allowed) && !wfi_in_wb && !fencei_in_wb etc.

Basically we want similar assertions for all states other than FUNCTIONAL (so also e.g. for SLEEP).

PR #705, opened 2022-11-28
Add rvfi_instr_prot and rvfi_mem_prot

silabs-oivind:


bhv/cv32e40x_rvfi_instr_obi.sv
```diff
@@ -53,21 +57,23 @@ module cv32e40x_rvfi_instr_obi import cv32e40x_pkg::*; import cv32e40x_rvfi_pkg:
...
+  assign response_valid = prefetch_resp_valid_i;
+  assign trans_accepted = prefetch_trans_valid_i && prefetch_trans_ready_i;
 
   //////////////////////////////////////////////////////////////////////////////
   // Count number of outstanding OBI transactions
   //////////////////////////////////////////////////////////////////////////////
 
-  assign outstanding_count_up   = m_c_obi_instr_if.s_req.req && m_c_obi_instr_if.s_gnt.gnt;
-  assign outstanding_count_down = m_c_obi_instr_if.s_rvalid.rvalid;
+  assign outstanding_count_up   = trans_accepted;
```
Silabs-ArjanB, line 75: Fix above comment (it is no longer counting outstanding OBI transactions). Also check if other comments referring to OBI need to be fixed.

bhv/cv32e40x_rvfi_instr_obi.sv
```diff
@@ -86,32 +92,19 @@ module cv32e40x_rvfi_instr_obi import cv32e40x_pkg::*; import cv32e40x_rvfi_pkg:
...
-
   // The RVFI instruction OBI FIFO tracks reads, writes and flushes of the alignment
   // buffer. As its size is at least the size of the alignment buffer, this FIFO
   // should never overflow. As the OBI response phase is at least a cycle after
   // the address phase acceptance, the address phase FIFO push will always be
   // to a different FIFO entry than the response phase FIFO push.
 
   // Push OBI address phase signals into FIFO when OBI address phase is accepted
-  assign fifo_req_push = m_c_obi_instr_if.s_req.req && m_c_obi_instr_if.s_gnt.gnt;
+  assign fifo_req_push = trans_accepted;
```
Silabs-ArjanB, line 104: It is not obvious anymore that the timing of the items pushed into the FIFO have the same timing as the fifo_req_push signal itself. It needs to be explained why it is okay to mix signals with seemingly different timing here.

docs/user_manual/source/rvfi.rst
```diff
@@ -223,11 +223,13 @@ For |corev|, the ``rvfi_mem`` interface has been expanded to support multiple me
    output [NRET * NMEM * XLEN/8 - 1 : 0] rvfi_mem_wmask
    output [NRET * NMEM * XLEN - 1 : 0]   rvfi_mem_rdata
    output [NRET * NMEM * XLEN - 1 : 0]   rvfi_mem_wdata
+   output [NRET * NMEM * 3    - 1 : 0]   rvfi_mem_prot
 
 Instructions will populate the ``rvfi_mem`` outputs with incrementing ``NMEM``, starting at ``NMEM=1``.
 
 Instructions with a single memory operation (e.g. all RV32I instructions), including split misaligned transfers, will only use NMEM = 1.
 Instructions with multiple memory operations (e.g. the push and pop instructions from Zcmp) use NMEM > 1 in case multiple memory operations actually occur.
+``rvfi_mem_prot`` indicates the value of OBI prot used for the memory access or accesses. Note that this will be undefined upon access faults.
```
Silabs-ArjanB, line 232: Next PRs keep documentation updates in a separate PR (that will result in a cleaner changelog)

sva/cv32e40x_load_store_unit_sva.sv
```diff
@@ -132,5 +133,25 @@ module cv32e40x_load_store_unit_sva
...
-endmodule // cv32e40x_load_store_unit_sva
 
+   // Helper logic to remember OBI prot for the first part of a split transfer
+  logic [2:0] trans_0_prot;
+  always_ff @(posedge clk, negedge rst_n) begin
+    if (!rst_n) begin
+      trans_0_prot <= '0;
+    end
+    else begin
+      if(m_c_obi_data_if.s_req.req && lsu_split_0_o) begin
```
Silabs-ArjanB, line 144: Pick signals that more obviously have the same timing

Silabs-ArjanB (changes requested):

bhv/cv32e40x_rvfi_instr_obi.sv
silabs-oivind: Done. Added some more explanation in the header as well.

bhv/cv32e40x_rvfi_instr_obi.sv
silabs-oivind: Added explanation.

docs/user_manual/source/rvfi.rst
silabs-oivind: Sorry, I'll keep that in mind.

sva/cv32e40x_load_store_unit_sva.sv
silabs-oivind: Updated this logic to only look at OBI req+gnt.

sva/cv32e40x_load_store_unit_sva.sv
```diff
@@ -132,5 +133,25 @@ module cv32e40x_load_store_unit_sva
...
+      if(m_c_obi_data_if.s_req.req && m_c_obi_data_if.s_gnt.gnt) begin
+        trans_prot_prev <= m_c_obi_data_if.req_payload.prot;
+      end
+    end
+  end
+
+  // Assert that OBI prot is equal for both parts of a split transfer
+  a_lsu_split_prot:
+    assert property (@(posedge clk) disable iff (!rst_n)
+                     (split_q && (m_c_obi_data_if.s_req.req && m_c_obi_data_if.s_gnt.gnt)) |->
```
Silabs-ArjanB, line 153: split_q does not have OBI timing. Did you check this assertion in the presence of bufferable stores?

Silabs-ArjanB (changes requested):

sva/cv32e40x_load_store_unit_sva.sv
silabs-oivind: Yes. This passes on both e40x and e40s with bufferable stores. 
However, closer inspection shows that split_q could be deasserted before the 2nd address phase when we have buffered stores (meaning prot from the first transfer would be compared to itself in this assertion). I updated the assertion to track split writes in a different way. Also added a comment describing this in the RTL.

sva/cv32e40x_load_store_unit_sva.sv
```diff
@@ -132,5 +133,50 @@ module cv32e40x_load_store_unit_sva
...
+        // Keep track of prot for the previous transfer
+        trans_prot_prev <= m_c_obi_data_if.req_payload.prot;
+      end
+
+      // Keep track of remaining parts of split transfers
+      if(ctrl_fsm_i.kill_ex) begin
+        // EX was killed, clear counter
+        split_rem_cnt <= 2'h0;
+      end
+      else if (lsu_split_0_o) begin
```
Silabs-ArjanB, line 157: This logic is still comparing signals with timing that is unrelated I think. I don't see why you can assume that lsu_split_0_o matches OBI req/gnt timing.

Silabs-ArjanB (changes requested):
Still the timing of signals is unclear/suspicious to me.

Also there is a merge conflict that needs to be resolved first.

Silabs-ArjanB (approved):

PR #707, opened 2022-11-29
Bugfix: LSU trigger match could cause tracking of split misaligned transactions to get out of sync

silabs-oysteink:
Suppressing LSU valid_0_i on trigger_match_0_i could cause the split_q flag to reset to zero before the last part of an LSU was allowed to go from EX to WB. This could cause an address change and change of the trigger match value.

Updated setting/resetting of split_q to take trigger matches into account.

Added assertions to check that split_q is always 0 when a new LSU instruction enters the EX stage and that the trigger match flag is stable while an LSU instruction remains in EX.

ed files:

### DIFF
--- a/bhv/cv32e40x_wrapper.sv
+++ b/bhv/cv32e40x_wrapper.sv
@@ -243,6 +243,9 @@ module cv32e40x_wrapper
       // The SVA's monitor modport can't connect to a master modport, so it is connected to the interface instance directly:
       .m_c_obi_data_if(core_i.m_c_obi_data_if),
       .ex_wb_pipe_i   (core_i.ex_wb_pipe),
+      .id_valid       (core_i.id_valid),
+      .ex_ready       (core_i.ex_ready),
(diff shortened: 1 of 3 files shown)


Silabs-ArjanB (changes requested):
I did not really like the introduction of the following line in the previous PR:

  assign valid_0_gated = valid_0_i && !trigger_match_0_i;

Now I see that it leads to further and further patching that should not have been required.  I think the introduction of valid_0_gated should be undone and the trigger match should be performed in a manner similar to how the MPU is treated, i.e. with a proper set of valid/ready signals and associated new transaction signals. Please update our docs first showing the relative position of the circuits handling trigger match, write buffer, MPU, etc. first.

silabs-oysteink:
Changes to LSU reverted, but kept the assertions for checking that split_q is always zero on the first part of a new LSU instruction (currently passes) and that trigger_match is always stable while a LSU instruction is in EX and WB stage is not yet ready (currently fails).

PR #724, opened 2022-12-13
Implement fence instruction

silabs-oivind:


silabs-oivind:
> It seems that fence can now get interrupted in the middle.

Ref our discussion, this is OK. Updated code and added comments as agreed upon.
SEC clean (with OBI assumes) to 59fb575922fed5e39d229b50f881429d5dbc0047

Silabs-ArjanB (approved):

Silabs-ArjanB (approved):

Issue #341, opened 2021-12-06
Sticky version of debug request


Silabs-ArjanB:
The debug request of the core is made sticky in two places it seems (i.e. in the debug module itself and in the core). That seems incorrect although the debug specification is quite unclear about haltreq as well (specifically where it states "this may cancel outstanding halt requests for those harts").

The following needs to be addressed:

- The 'this may' part of the debug specification needs to get clarified. For the 40X the core itself will also remember the debug request, which offers less controller to the debugger.
- Any debug entry (i.e. not only debug entry due to a debug request) will cause debug_req_q to get cleared, which seems like an issue as well (see https://github.com/openhwgroup/cv32e40x/blob/af71c0ec7754fd87ce25335ef584de605aa68ca9/rtl/cv32e40x_controller_fsm.sv#L742)

Silabs-ArjanB:
See clarifications at https://github.com/riscv/riscv-debug-spec/issues/753

Silabs-ArjanB:
Can be closed once user manual has been updated as well.

PR #725, opened 2022-12-13
Fix for issue #341.

silabs-oysteink:
Removed sticky debug request (debug_req_q)

Updated assertion a_fencei_hndshk_req_when_fencei_wb to accomodate changes.

### DIFF
--- a/rtl/cv32e40x_controller_fsm.sv
+++ b/rtl/cv32e40x_controller_fsm.sv
@@ -38,7 +38,6 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
 (
   // Clocks and reset
   input  logic        clk,                        // Gated clock
-  input  logic        clk_ungated_i,              // Ungated clock
   input  logic        rst_n,
 
   input  logic        fetch_enable_i,             // Start executing
--- a/rtl/cv32e40x_controller_fsm.sv
+++ b/rtl/cv32e40x_controller_fsm.sv
@@ -137,9 +137,6 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
   // Debug state
   debug_state_e debug_fsm_cs, debug_fsm_ns;
 
-  // Sticky version of debug_req_i
-  logic debug_req_q;
-
   // Sticky version of lsu_err_wb_i
   logic nmi_pending_q;
   logic nmi_is_store_q; // 1 for store, 0 for load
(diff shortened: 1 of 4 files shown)


sva/cv32e40x_controller_fsm_sva.sv
```diff
@@ -244,9 +244,10 @@ module cv32e40x_controller_fsm_sva
       else `uvm_error("controller", "Fencei request when no fencei in writeback")
 
   // Assert that the fencei request is set the cycle after fencei instruction enters WB (if fencei_ready=1 and there are no higher priority events)
+  // Only check when no higher priority event is pending (nmi, async debug or interrupts) and WB stage is not killed
```
Silabs-ArjanB, line 247: Can you explain why this assertion was required to be changed?

Silabs-ArjanB (changes requested):

sva/cv32e40x_controller_fsm_sva.sv
silabs-oysteink: When debug_req_q was still in the controller, it would remain set until we actually entered debug mode (causing pending_async_debug to be 1). That would cause the antecedent of the assertion not to trig. Now that the sticky debug is gone, we may go to DEBUG_TAKEN due to a short pulse on debug_req_i and pending_async_debug may be 0 while in the DEBUG_TAKEN state. We are then killing the WB stage, but the antecedent didn't take killing of WB into account and only looked at signals originating from the ex_wb_pipe.

rtl/cv32e40x_controller_fsm.sv
```diff
@@ -482,8 +479,12 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
...
                               (ebreak_in_wb && debug_mode_q); // Ebreak during debug_mode restarts execution from dm_halt_addr, as a regular debug entry without CSR updates.
 
-  // Debug pending for external debug request
-  assign pending_async_debug = ((debug_req_i || debug_req_q) && !debug_mode_q);
+  // Debug pending for external debug request, only if not already in debug mode
+  // Ideally the !debug_mode_q below should be factored into async_debug_allowed, but
+  // that can currently cause a deadlock if debug_req_i gets asserted while in debug mode, as
+  // a pending but not allowed async debug will cause the ID stage to halt forever while trying
+  // to get to an interruptible state.
+  assign pending_async_debug = debug_req_i && !debug_mode_q;
```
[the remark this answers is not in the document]
silabs-oysteink: Removed ungated clock from controller.

Silabs-ArjanB (approved):

PR #733, opened 2022-12-14
Maded tdata1 trigger type 0xF the default for WARL resolution.

silabs-oysteink:


Silabs-ArjanB (approved):

Silabs-ArjanB:
This change is needed as part of the required behavior upon writing 0x0 to tdata1

PR #738, opened 2022-12-20
Updated aliasing of mpp and mpie between mcause and mstatus.

silabs-oysteink:
RVFI didn't always get the correct write values, and the aliasing was only working from mcause to mstatus and not the other way.

The three bits are now duplicated in both mcause and mstatus, and are always written simultaneously. Assertions are added to check that mcause and mstatus always write at the same time, and that the *_rdata.mpp/mpie fields are always the same.

SEC clean

### DIFF
--- a/rtl/cv32e40x_cs_registers.sv
+++ b/rtl/cv32e40x_cs_registers.sv
@@ -735,7 +735,7 @@ module cv32e40x_cs_registers import cv32e40x_pkg::*;
       mcause_n                 = '{
                                     irq:            csr_wdata_int[31],
                                     minhv:          csr_wdata_int[30],
-                                    mpp:            mstatus_mpp_resolve(mstatus_rdata.mpp, csr_wdata_int[MCAUSE_MPP_BIT_HIGH:MCAUSE_MPP_BIT_LOW]),
+                                    mpp:            mcause_mpp_resolve(mcause_rdata.mpp, csr_wdata_int[MCAUSE_MPP_BIT_HIGH:MCAUSE_MPP_BIT_LOW]),
                                     mpie:           csr_wdata_int[MCAUSE_MPIE_BIT],
                                     mpil:           csr_wdata_int[23:16],
                                     exception_code: csr_wdata_int[10:0],
@@ -880,12 +880,15 @@ module cv32e40x_cs_registers import cv32e40x_pkg::*;
           if (SMCLIC) begin
             mnxti_we = 1'b1;
 
-            // Writes to mnxti also writes to mstatus
+            // Writes to mnxti also writes to mstatus (uses mstatus in the RMW operation)
+            // Also writing to mcause to ensure we can assert mstatus_we == mcause_we and similar for mpp/mpie.
             mstatus_we = 1'b1;
             mcause_we  = 1'b1;
 
             // Writes to mintstatus.mil and mcause depend on the current state of
             // clic interrupts AND the type of CSR instruction used.
+            // Mcause is written unconditionally for aliasing purposes, but the mcause_n
+            // is modified to reflect the side effects in case mnxti_irq_pending_i i set.
             // Side effects occur when there is an actual write to the mstatus CSR
             // This is already coded into the csr_we_int/mnxti_we
             if (mnxti_irq_pending_i) begin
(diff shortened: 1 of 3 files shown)


rtl/cv32e40x_cs_registers.sv
```diff
@@ -1459,7 +1521,7 @@ module cv32e40x_cs_registers import cv32e40x_pkg::*;
   // Signal when an interrupt may become enabled due to a CSR write
   generate
     if (SMCLIC) begin : smclic_irq_en
-      assign csr_irq_enable_write_o = mstatus_we || priv_lvl_we || mintthresh_we || mintstatus_we;
+      assign csr_irq_enable_write_o = (mstatus_we && !ctrl_fsm_i.csr_clear_minhv) || priv_lvl_we || mintthresh_we || mintstatus_we;
```
silabs-oysteink, line 1524: Needed to be SEC clean. Not needed functionally, as a clearing of minhv will only touch that bit and never update any interrupt enables. Could keep it as it was, or modify to not raise the signal on write to mstatus where we know that mie will not be written?

rtl/cv32e40x_cs_registers.sv
```diff
@@ -1337,6 +1391,19 @@ module cv32e40x_cs_registers import cv32e40x_pkg::*;
 
     end else begin : basic_mode_csrs
 
+      cv32e40x_csr
+      #(
+        .WIDTH      (32),
+        .RESETVALUE (32'd0)
```
Silabs-ArjanB, line 1397: Introduce MCAUSE_BASIC_RESET_VAL and MCAUSE_CLIC_RESET_VAL (just like for MTVEC)

rtl/cv32e40x_cs_registers.sv
Silabs-ArjanB: I prefer the non-SEC clean variant. The only negative effect could be an unneeded stall, right?

sva/cv32e40x_cs_registers_sva.sv
```diff
@@ -135,6 +139,29 @@ module cv32e40x_cs_registers_sva
...
+    property p_mstatus_mcause_we;
+      @(posedge clk) disable iff (!rst_n)
+      (
+        1'b1
+        |->
+        mstatus_we == mcause_we
+      );
+    endproperty;
+    a_mstatus_mcause_we: assert property(p_mstatus_mcause_we)
+      else `uvm_error("cs_registers", "Mstatus.mpp and mstatus.mpie not written at the same time")
```
Silabs-ArjanB, line 151: Fix error message

Silabs-ArjanB (changes requested):

rtl/cv32e40x_cs_registers.sv
silabs-oysteink: Yes, it would cause an extra stall, but I believe the scenario creating this isn't very common. I will remove the non-SEC code in my final push.

rtl/cv32e40x_cs_registers.sv
```diff
@@ -725,11 +735,14 @@ module cv32e40x_cs_registers import cv32e40x_pkg::*;
       mcause_n                 = '{
                                     irq:            csr_wdata_int[31],
                                     minhv:          csr_wdata_int[30],
+                                    mpp:            mstatus_mpp_resolve(mstatus_rdata.mpp, csr_wdata_int[MCAUSE_MPP_BIT_HIGH:MCAUSE_MPP_BIT_LOW]),
```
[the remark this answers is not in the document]
silabs-oysteink: Done

rtl/cv32e40x_cs_registers.sv
silabs-oysteink: Removed check for clear_minhv, resulting PR is not SEC clean.

rtl/cv32e40x_cs_registers.sv
```diff
@@ -1098,6 +1142,10 @@ module cv32e40x_cs_registers import cv32e40x_pkg::*;
         mstatus_n.mprv = (privlvl_t'(dcsr_rdata.prv) == PRIV_LVL_M) ? mstatus_rdata.mprv : 1'b0;
         mstatus_we     = 1'b1;
 
+        // Not really needed, but allows for asserting mstatus_we == mcause_we to check aliasing formally
```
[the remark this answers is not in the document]
silabs-oysteink: Done

rtl/cv32e40x_cs_registers.sv
```diff
@@ -1037,7 +1071,15 @@ module cv32e40x_cs_registers import cv32e40x_pkg::*;
           mepc_n         = ctrl_fsm_i.pipe_pc;
           mepc_we        = 1'b1;
 
-          mcause_n       = ctrl_fsm_i.csr_cause;
+          // Save relevant fields from controller to mcause
```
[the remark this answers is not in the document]
silabs-oysteink: Moved relevant fields into "if (SMCLIC)"

rtl/cv32e40x_cs_registers.sv
silabs-oysteink: Done

sva/cv32e40x_cs_registers_sva.sv
silabs-oysteink: Fixed

Silabs-ArjanB (approved):

Issue #668, opened 2022-09-22
Additional feedback on https://github.com/openhwgroup/cv32e40x/pull/661


Silabs-ArjanB:
Please see https://github.com/openhwgroup/cv32e40x/pull/661 for additional feedback on the already merged pull request.

silabs-oysteink:
This should have been fixed with PR #741.

PR #741, opened 2022-12-20
Fix for issue #668

silabs-oysteink:
Fixed some feedback that was previously not incorporated. 

Added possibility to control the RVFI _rmask values. These will always be all ones for most CSRs, but for mscratchcsw[l] and mnxti the rmasks will only be set whenever the according CSR is actually accessed in WB.

### DIFF
--- a/rtl/cv32e40x_cs_registers.sv
+++ b/rtl/cv32e40x_cs_registers.sv
@@ -280,6 +280,11 @@ module cv32e40x_cs_registers import cv32e40x_pkg::*;
   logic [31:0]                  mhpmcounter_write_upper;                        // Write 32 upper bits mhpmcounter_q
   logic [31:0]                  mhpmcounter_write_increment;                    // Write increment of mhpmcounter_q
 
+  // Signal used for RVFI to set rmask, not used internally
+  logic                         mscratchcsw_in_wb;
+  logic                         mscratchcswl_in_wb;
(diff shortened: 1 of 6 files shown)


rtl/cv32e40x_cs_registers.sv
```diff
@@ -1890,10 +1895,16 @@ module cv32e40x_cs_registers import cv32e40x_pkg::*;
 
   assign mcountinhibit_rdata = mcountinhibit_q;
 
+  // Assign values used for setting rmask in RVFI
+  assign mscratchcsw_in_wb = (csr_waddr == CSR_MSCRATCHCSW);
```
Silabs-ArjanB, line 1899: These signals can now remain asserted even if there is no CSR instruction in WB at all. Also add a comment in EX that csr_waddr is used even for instructions that do not write the CSR (so that we do not clock gate these signals further).

Silabs-ArjanB (changes requested):

rtl/cv32e40x_cs_registers.sv
silabs-oysteink: Added ex_wb_pipe.csr_en as qualifier, and updated comment in EX stage about the ex_wb_pipe.csr_addr

Silabs-ArjanB (approved):

Issue #410, opened 2022-01-28
Transition from debug_havereset to debug_halted not possible


silabs-oysteink:
The debug state fsm currently does not allow the transition from debug_havereset to debug_halted (see image).
![image](https://user-images.githubusercontent.com/66771756/151547288-cddb8b95-6b9e-4f8d-8f73-ab7a85e401d9.png)

The controller needs an update to ensure the transition will be possible. If debug_req_i is asserted early, the core will set debug_havereset -> debug_running -> debug_halted. Possible solution is to check for debug entry in the state BOOT_SET, and se debug_mode_n accordingly.

Silabs-ArjanB:
Fixed by https://github.com/openhwgroup/cv32e40x/pull/747

PR #747, opened 2022-12-22
Fix for issue #410

silabs-oysteink:
Removed BOOT_SET state from controller_fsm.
Checking for debug_req_i in RESET state to enable going directly to debug mode from RESET without fetching any instructions.

### DIFF
--- a/rtl/include/cv32e40x_pkg.sv
+++ b/rtl/include/cv32e40x_pkg.sv
@@ -169,7 +169,7 @@ typedef enum logic [DIV_OP_WIDTH-1:0]
  } div_opcode_e;
 
 // FSM state encoding
-typedef enum logic [2:0] { RESET, FUNCTIONAL, SLEEP, DEBUG_TAKEN} ctrl_state_e;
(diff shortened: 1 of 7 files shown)


rtl/include/cv32e40x_pkg.sv
```diff
@@ -169,7 +169,7 @@ typedef enum logic [DIV_OP_WIDTH-1:0]
  } div_opcode_e;
 
 // FSM state encoding
-typedef enum logic [2:0] { RESET, BOOT_SET, FUNCTIONAL, SLEEP, DEBUG_TAKEN} ctrl_state_e;
+typedef enum logic [2:0] { RESET, FUNCTIONAL, SLEEP, DEBUG_TAKEN} ctrl_state_e;
```
Silabs-ArjanB, line 172: Only 2 bits needed now

Silabs-ArjanB (changes requested):

rtl/include/cv32e40x_pkg.sv
silabs-oysteink: Changed to two bits

Silabs-ArjanB (approved):

PR #748, opened 2023-01-02
Add readme to yaml directory

silabs-robin:


silabs-robin:
@silabs-oivind, please let me know if this sufficiently describes how to make use of the csr yaml file or if a more detailed explanation should be added.

Silabs-ArjanB (changes requested):
I think it would be better if the reference to 40s are replaced by 40x in this file

silabs-robin:
> I think it would be better if the reference to 40s are replaced by 40x in this file

Right, makes sense. How would that be kept in sync when merging between the X and S repo? Is there a neutral way to write it? Must it always be manually kept in sync? Will the files simply be different?

silabs-robin:
> > > I think it would be better if the reference to 40s are replaced by 40x in this file
> > 
> > Right, makes sense. How would that be kept in sync when merging between the X and S repo?
> 
> After merging X->S we need to rename e40x->e40s in the cv32e40s repo. Subsequent merges from X to S should be unproblematic.

Then that part of it is ok.

Another problem: The two links have no X equivalent yet. Both the preproc script and the readme is on the cv32e40s/dev branch only atm. So changing the links now would just lead to 404.
Do we accept temporary 404 in anticipation of a future cvverif merge? Do we hold off merging this readme until the next cvverif s<->x merge? Do we accept that a 40x readme file links to 40s documents?

Silabs-ArjanB:
> 

Just fix the 404 later

silabs-robin:
> I think it would be better if the reference to 40s are replaced by 40x in this file

> Just fix the 404 later

Fixed.

Silabs-ArjanB (approved):

PR #749, opened 2023-01-02
Fix for issues #365 and #361 on cv32e40s.

silabs-oysteink:
When waking up from sleep (exiting FSM state SLEEP), the cause of wakeup is now flopped to enable the FSM to take the action according to the wakeup reason, although there may be higher priority events appearing after the wakeup.

### DIFF
--- a/sva/cv32e40x_controller_fsm_sva.sv
+++ b/sva/cv32e40x_controller_fsm_sva.sv
@@ -112,7 +112,11 @@ module cv32e40x_controller_fsm_sva
   input logic           debug_req_i,
   input logic           fetch_enable_i,
   input logic           instr_req_o,
-  input logic           instr_dbg_o
+  input logic           instr_dbg_o,
+  input logic           wfe_in_wb,
+  input mstatus_t       mstatus_i,
+  input logic           woke_to_interrupt_q,
+  input logic           woke_to_debug_q
 );
 
 
@@ -288,9 +292,11 @@ module cv32e40x_controller_fsm_sva
       else `uvm_error("controller", "Fencei handshake active while lsu_busy_o = 1")
 
   // assert that NMI's are not reported on irq_ack
+  // Exception for the case where the core wakes from SLEEP due to an interrupt
+  //   - in that case the interrupt is honored while there may be a pending nmi.
   a_irq_ack_no_nmi :
     assert property (@(posedge clk) disable iff (!rst_n)
-                     ctrl_fsm_o.irq_ack |-> !pending_nmi)
+                     ctrl_fsm_o.irq_ack |-> !(pending_nmi && !woke_to_interrupt_q))
       else `uvm_error("controller", "irq_ack set while there's a pending NMI")
 
   // Assert that intr_taken is always single cycle. I.e. no double counting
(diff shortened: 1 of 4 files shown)


sva/cv32e40x_controller_fsm_sva.sv
```diff
@@ -875,8 +881,52 @@ end
...
     else `uvm_error("controller", "Debug not entered on a WPT match")
 
-
-
+  // Ensure debug mode is entered if woken up by a debug request
+  a_sleep_to_debug:
+  assert property (@(posedge clk) disable iff (!rst_n)
+                  (ctrl_fsm_cs == SLEEP) &&
+                  (ctrl_fsm_ns == FUNCTIONAL) &&
+                  !(pending_nmi || irq_wu_ctrl_i || (wfe_in_wb && wu_wfe_i))
```
Silabs-ArjanB, line 889: Wouldn't it be logical to add && debug_req_i here?

sva/cv32e40x_controller_fsm_sva.sv
```diff
@@ -875,8 +881,52 @@ end
...
+                  (ctrl_fsm_ns == DEBUG_TAKEN))
+    else `uvm_error("controller", "Woke from sleep due to debug_req but debug mode not entered")
+
+  // Ensure interrupt is taken if woken up by an interrupt
+  a_sleep_to_irq:
+  assert property (@(posedge clk) disable iff (!rst_n)
+                  (ctrl_fsm_cs == SLEEP) &&
+                  (ctrl_fsm_ns == FUNCTIONAL) &&
+                  !(pending_nmi || debug_req_i || (wfe_in_wb && wu_wfe_i)) &&
+                  mstatus_i.mie
```
Silabs-ArjanB, line 900: && irq_wu_ctrl_i seems more logical than the currently used && mstatus_i.mie

sva/cv32e40x_controller_fsm_sva.sv
```diff
@@ -875,8 +881,52 @@ end
...
+                  |=>
+                  (ctrl_fsm_o.irq_ack))
+    else `uvm_error("controller", "Woke from sleep due to irq but irq not taken")
+
+  // Ensure NMI is taken if woken up by an NMI
+  a_sleep_to_nmi:
+  assert property (@(posedge clk) disable iff (!rst_n)
+                  (ctrl_fsm_cs == SLEEP) &&
+                  (ctrl_fsm_ns == FUNCTIONAL) &&
+                  !(debug_req_i || irq_wu_ctrl_i || (wfe_in_wb && wu_wfe_i))
```
Silabs-ArjanB, line 910: Add && pending_nmi

sva/cv32e40x_controller_fsm_sva.sv
```diff
@@ -875,8 +881,52 @@ end
...
+
+  // Ensure interrupt is taken if woken up by an interrupt
+  a_sleep_to_irq:
+  assert property (@(posedge clk) disable iff (!rst_n)
+                  (ctrl_fsm_cs == SLEEP) &&
+                  (ctrl_fsm_ns == FUNCTIONAL) &&
+                  !(pending_nmi || debug_req_i || (wfe_in_wb && wu_wfe_i)) &&
+                  mstatus_i.mie
+                  |=>
+                  (ctrl_fsm_o.irq_ack))
```
Silabs-ArjanB, line 902: Use pc_mux instead to make it similar to next assertion

Silabs-ArjanB (changes requested):
Only some minor suggestions to make the added assertions look more similar

sva/cv32e40x_controller_fsm_sva.sv
silabs-oysteink: Agreed, adding it

sva/cv32e40x_controller_fsm_sva.sv
silabs-oysteink: Agree that it is logical to add the '&& irq_wu_ctrl_i'. mstatus.mie is still needed for this assertion as the core may wake up on a non-globally enabled interrupt. In that case no interrupt will be taken and the core continues ad PC+4.

sva/cv32e40x_controller_fsm_sva.sv
silabs-oysteink: Will add it

sva/cv32e40x_controller_fsm_sva.sv
silabs-oysteink: Changing to pc_mux comparisons.

Silabs-ArjanB (approved):

PR #769, opened 2023-01-25
Bugfix for wpt/mpu backpressure from WB

silabs-oysteink:
Bugfix for wpt/mpu backpressure from WB:
    
- resp_valid could be high for several cycles if WB was halted (debug entry or XIF). This would cause the LSU outstanding counter to underflow.
- Fixed by introducing 'resp_ready' in the LSU, which is connected to wb_ready via ready_0_i. This causes the outstanding counter to only decrement when there is a valid response _and_ WB is ready. 
- Added assertion to check that WB must be ready when a LSU response arrives from the bus.

### DIFF
--- a/bhv/cv32e40x_wrapper.sv
+++ b/bhv/cv32e40x_wrapper.sv
@@ -188,6 +188,7 @@ module cv32e40x_wrapper
   bind cv32e40x_wb_stage:
     core_i.wb_stage_i cv32e40x_wb_stage_sva wb_stage_sva
     (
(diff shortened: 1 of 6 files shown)


Silabs-ArjanB (changes requested):
It is difficult to see if this actually works or not because not all resp_valid signals are treated the same. I think we need to consider the high level approach to this issue. Also notice that the original back pressure issue has not been solved either (signals can still change too early in WB.

silabs-oysteink:
Closing, as the whole solution with backpressure will be reverted for a better solution.

PR #772, opened 2023-01-30
PUSH/POP not allowed outside of 'main' PMA regions

silabs-oysteink:


rtl/cv32e40x_sequencer.sv
```diff
@@ -170,6 +172,7 @@ module cv32e40x_sequencer import cv32e40x_pkg::*;
               end
             end
             3'b110: begin
+              seq_pushpop_o = 1'b1;
```
Silabs-ArjanB, line 175: I prefer if illegal isntructions are excluded from having seq_push_pop_o set to 1. Same remark for line 191

Silabs-ArjanB (changes requested):

rtl/cv32e40x_sequencer.sv
silabs-oysteink: Agreed, fixed.

Silabs-ArjanB (approved):

PR #785, opened 2023-02-16
Split A_EXT parameter and added wu_wrs_i pin.

silabs-oysteink:
Split A_EXT parameter into A_NONE, ZALRSC and A. Updated a_decoder according t othe new split A_EXT parameter.

Added wu_wrs_i pin, currently not used.

### DIFF
--- a/rtl/cv32e40x_cs_registers.sv
+++ b/rtl/cv32e40x_cs_registers.sv
@@ -120,7 +120,7 @@ module cv32e40x_cs_registers import cv32e40x_pkg::*;
 );
 
   localparam logic [31:0] CORE_MISA =
-    (32'(A_EXT != A_NONE) <<  0) | // A - Atomic Instructions extension
+    (32'(A_EXT == A)      <<  0) | // A - Atomic Instructions extension
     (32'(1)               <<  2) | // C - Compressed extension
     (32'(RV32 == RV32E)   <<  4) | // E - RV32E/64E base ISA
     (32'(RV32 == RV32I)   <<  8) | // I - RV32I/64I/128I base ISA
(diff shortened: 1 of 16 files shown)


rtl/cv32e40x_cs_registers.sv
```diff
@@ -120,14 +120,14 @@ module cv32e40x_cs_registers import cv32e40x_pkg::*;
...
   localparam logic [31:0] CORE_MISA =
-    (32'(A_EXT)         <<  0) | // A - Atomic Instructions extension
-    (32'(1)             <<  2) | // C - Compressed extension
-    (32'(RV32 == RV32E) <<  4) | // E - RV32E/64E base ISA
-    (32'(RV32 == RV32I) <<  8) | // I - RV32I/64I/128I base ISA
-    (32'(M_EXT == M)    << 12) | // M - Integer Multiply/Divide extension
-    (32'(0)             << 20) | // U - User mode implemented
-    (32'(1)             << 23) | // X - Non-standard extensions present
-    (32'(MXL)           << 30);  // M-XLEN
+    (32'(A_EXT != A_NONE) <<  0) | // A - Atomic Instructions extension
```
Silabs-ArjanB, line 123: I 'think' this should be A_EXT == A instead (i.e. not setting the misa.A bit when only supporting part of the A extension.

Silabs-ArjanB (changes requested):

rtl/cv32e40x_cs_registers.sv
silabs-oysteink: Agreed, I will change it.

Silabs-ArjanB (approved):

PR #786, opened 2023-02-16
RVFI updates for A_EXT

silabs-oysteink:
Misaligned atomic instruction will get rvfi_trap.cause_type == 0
Aligned atomic instructions to non-atomic PMA regions will get rvfi_trap.cause_type == 1

Note that AMOs are not properly supported within RVFI  yet. This will be implemented once verification is closer to start looking at AMOs.

### DIFF
--- a/bhv/cv32e40x_rvfi.sv
+++ b/bhv/cv32e40x_rvfi.sv
@@ -22,9 +22,9 @@ module cv32e40x_rvfi
   import cv32e40x_pkg::*;
   import cv32e40x_rvfi_pkg::*;
   #(
-    parameter bit SMCLIC = 0,
-    parameter int DEBUG  = 1,
-    parameter bit A_EXT  = 0
+    parameter bit     SMCLIC = 0,
+    parameter int     DEBUG  = 1,
+    parameter a_ext_e A_EXT  = A_NONE
   )
   (
    input logic                                clk_i,
@@ -81,6 +81,7 @@ module cv32e40x_rvfi
    input logic                                lsu_pmp_err_ex_i,
    input logic                                lsu_pma_err_ex_i,
    input logic                                lsu_pma_atomic_ex_i,
+   input logic                                lsu_misaligned_ex_i,
    input obi_data_req_t                       buffer_trans,
    input logic                                lsu_split_q_ex_i,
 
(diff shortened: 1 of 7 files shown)


bhv/cv32e40x_rvfi.sv
```diff
@@ -702,8 +705,15 @@ module cv32e40x_rvfi
...
   logic         mret_ptr_wb;
 
+  // Detect a PMA error due to atomics
+  logic         lsu_pma_err_atomic_ex;
+
   assign        mret_ptr_wb = mret_ptr_wb_i;
 
+  // PMA error due to atomic not within an atomic region
+  // Error due to misaligned atomics shall get cause_type==0 and is excluded here.
+  assign lsu_pma_err_atomic_ex = lsu_pma_err_ex_i && lsu_pma_atomic_ex_i && !lsu_misaligned_ex_i;
```
Silabs-ArjanB, line 715: This should not be needed

bhv/cv32e40x_rvfi.sv
```diff
@@ -1088,7 +1098,7 @@ module cv32e40x_rvfi
         end
 
         mem_err   [STAGE_WB]  <= lsu_pmp_err_ex_i        ? MEM_ERR_PMP :
-                                 lsu_pma_err_atomic_ex_i ? MEM_ERR_ATOMIC :
+                                 lsu_pma_err_atomic_ex   ? MEM_ERR_ATOMIC :
```
Silabs-ArjanB, line 1101: If this statement was written in the order of priorities (now also PMP error seemingly get highest priority) (might not actually be the case, but difficult to see from this code)

Silabs-ArjanB (changes requested):

bhv/cv32e40x_rvfi.sv
silabs-oysteink: Agreed

bhv/cv32e40x_rvfi.sv
silabs-oysteink: Introduced a fourth cause_type for misaligned atomics as discussed and rewrote priority logic.

Silabs-ArjanB (approved):

PR #787, opened 2023-02-16
User manual update for A_EXT.

silabs-oysteink:
Updated user manual to describe the updated A_EXT parameter, and added a note about mret instructions not clearing reservation sets.


Silabs-ArjanB (approved):
Made some small changes

PR #794, opened 2023-02-27
Added exception codes 0x4 and 0x6 misaligned atomics.

silabs-oysteink:
These will only be used for misaligned atomics which are not otherwise blocked by the PMA.

SEC clean when A_EXT=A_NONE and the new exception codes are not allowed to cause exception triggers (tdata2_rdata may propagate and lead to counter examples).

### DIFF
--- a/rtl/cv32e40x_mpu.sv
+++ b/rtl/cv32e40x_mpu.sv
@@ -119,13 +119,13 @@ module cv32e40x_mpu import cv32e40x_pkg::*;
           if (core_mpu_err_wait_i) begin
             if(core_trans_we) begin
               // MPU error on write
-              // PMA errors take presedence over misaligned atomics
+              // PMA errors take precedence over misaligned atomics
               state_n = core_one_txn_pend_n ? (pma_err ? MPU_WR_ERR_RESP : MPU_WR_MISALIGN_RESP) :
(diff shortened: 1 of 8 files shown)


rtl/cv32e40x_mpu.sv
```diff
@@ -118,35 +119,47 @@ module cv32e40x_mpu import cv32e40x_pkg::*;
...
               // MPU error on write
-              state_n = core_one_txn_pend_n ? MPU_WR_ERR_RESP : MPU_WR_ERR_WAIT;
+              // PMA errors take presedence over misaligned atomics
+              state_n = core_one_txn_pend_n ? (pma_err ? MPU_WR_ERR_RESP : MPU_WR_MISALIGN_RESP) :
+                                              (pma_err ? MPU_WR_ERR_WAIT : MPU_WR_MISALIGN_WAIT);
             end
             else begin
               // MPU error on read
-              state_n = core_one_txn_pend_n ? MPU_RE_ERR_RESP : MPU_RE_ERR_WAIT;
+              // PMA errors take presedence over misaligned atomics
```
Silabs-ArjanB, line 128: presedence -> precedence

Silabs-ArjanB (changes requested):

Silabs-ArjanB (approved):

PR #806, opened 2023-03-16
Fix for cv32e40s issue 415

silabs-oivind:
Fix for https://github.com/openhwgroup/cv32e40s/issues/415
SEC clean, except for debug_pc_o.


Silabs-ArjanB (commented):
The change looks good. The comment in the PR however states that this is a SEC clean change. That cannot be true if it is also a bug fix. Can you please explain.

silabs-oivind:
@Silabs-ArjanB. You're right! It's SEC clean for all signals but debug_pc_o.

PR #812, opened 2023-03-22
Misaligned pointer update.

silabs-oysteink:
Detecting misaligned pointers in the IF stage. Propagating alignment status down the pipeline and flagging exception code 0x0 if not ok.

Added assertions for checking that only mret pointers can cause misaligned instruction address exceptions. Added assertions to check that the alignment buffer only signal one valid for each pointer (currently failing), and that the alignment buffer must signal instr_valid_o if is not currently prefetching and the controller wants us to prefetch (currently failing).

The new and failing assertions will be addressed in a coming PR

### DIFF
--- a/rtl/cv32e40x_alignment_buffer.sv
+++ b/rtl/cv32e40x_alignment_buffer.sv
@@ -562,6 +562,11 @@ module cv32e40x_alignment_buffer import cv32e40x_pkg::*;
         if(instr_valid_o && instr_ready_i) begin
           addr_q <= addr_n;
           rptr   <= rptr_n;
+
+          // Clear pointer flags when pointers are consumed.
+          is_clic_ptr_q     <= 1'b0;
+          is_mret_ptr_q     <= 1'b0;
+          is_tbljmp_ptr_q   <= 1'b0;
         end
       end
 
--- a/rtl/cv32e40x_alignment_buffer.sv
+++ b/rtl/cv32e40x_alignment_buffer.sv
@@ -199,7 +199,7 @@ module cv32e40x_alignment_buffer import cv32e40x_pkg::*;
           mpu_status_unaligned = MPU_RE_FAULT;
         end
 
-        if((resp_q[rptr2].align_status != MPU_OK) || (resp_q[rptr].align_status != MPU_OK)) begin
+        if((resp_q[rptr2].align_status != ALIGN_OK) || (resp_q[rptr].align_status != ALIGN_OK)) begin
           align_status_unaligned = ALIGN_RE_ERR;
         end
 
@@ -224,7 +224,7 @@ module cv32e40x_alignment_buffer import cv32e40x_pkg::*;
             mpu_status_unaligned = MPU_RE_FAULT;
           end
 
-          if((resp_q[rptr].align_status != MPU_OK) || (resp_i.align_status != MPU_OK)) begin
+          if((resp_q[rptr].align_status != ALIGN_OK) || (resp_i.align_status != ALIGN_OK)) begin
             align_status_unaligned = ALIGN_RE_ERR;
           end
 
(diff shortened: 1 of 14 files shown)


rtl/cv32e40x_alignment_buffer.sv
```diff
@@ -214,19 +224,24 @@ module cv32e40x_alignment_buffer import cv32e40x_pkg::*;
             mpu_status_unaligned = MPU_RE_FAULT;
           end
 
+          if((resp_q[rptr].align_status != MPU_OK) || (resp_i.align_status != MPU_OK)) begin
```
silabs-oivind, line 227: MPU_OK -> ALIGN_OK

rtl/cv32e40x_controller_fsm.sv
silabs-oivind, line 1: align_status should be checked wherever mpu_status is checked. Seems to be missing on line 1028:
            if (!(if_id_pipe_i.instr.bus_resp.err || (if_id_pipe_i.instr.mpu_status != MPU_OK))) begin // todo: add check for alignment check in IF (mret pointer)

silabs-oivind (changes requested):
Please see comments in files.

silabs-oysteink:
Added four new commits:

1: Clearing of 'instr_is_*_ptr_o' flags in the alignment buffer. This ensure the IF stage does not see any pointer flags after a pointer has been sent from IF to ID. Not a real issue on E40X, but on E40S this could cause dummy instructions to be marked as pointers. SEC state unknown as it runs for a really long time.
2: Fixes according to review comments so far - not SEC clean as a check of align_status within the controller was added.
3: Fixed the issue where pointers could issue two instr_valid_o for the cases where the pointer itself looked like two compressed instructions. - Not SEC clean (bugfix)
4: Fixed an issue where the alignment buffer would deadlock for a misaligned pointer (it expected two responses). (Not SEC clean, bugfix)

The assertion 'a_no_outstanding_instr_valid attempts to check for deadlocks within the alignment buffer. If there are no outstanding requests or new requests being initiated, the alignment buffer must signal a valid instruction on instr_valid_o.

rtl/cv32e40x_alignment_buffer.sv
```diff
@@ -178,13 +181,19 @@ module cv32e40x_alignment_buffer import cv32e40x_pkg::*;
...
   // unaligned_is_compressed and aligned_is_compressed are only defined when valid = 1 (which implies that instr_valid_o will be 1)
-  assign unaligned_is_compressed = instr[17:16] != 2'b11;
-  assign aligned_is_compressed   = instr[1:0] != 2'b11;
+  // Never flag a compressed instruction (aligned or misaligned) for pointers. Otherwise one could signal instr_valid_o twice for one pointer
+  // if the pointer itself looks like two compressed instructions.
+  assign unaligned_is_compressed = (instr[17:16] != 2'b11) &&
+                                   !(instr_is_clic_ptr_o || instr_is_mret_ptr_o || instr_is_tbljmp_ptr_o);
+
+  assign aligned_is_compressed   = (instr[1:0] != 2'b11) &&
+                                   !(instr_is_clic_ptr_o || instr_is_mret_ptr_o || instr_is_tbljmp_ptr_o);
```
silabs-oysteink, line 190: Not setting these signals for pointers ensure we never issue two instr_valid_o for pointers.

rtl/cv32e40x_alignment_buffer.sv
```diff
@@ -248,14 +269,20 @@ module cv32e40x_alignment_buffer import cv32e40x_pkg::*;
...
+      instr_instr_o.align_status   = align_status_unaligned;
+
       if (!valid) begin
+        // No instruction valid
         instr_valid_o = 1'b0;
-      // Unaligned instruction is compressed, we only need 16 upper bits from index 0
+      end else if (instr_is_clic_ptr_o || instr_is_mret_ptr_o || instr_is_tbljmp_ptr_o) begin
+        // currently outputting a pointer, valid whenever the response arrives or we have
+        // the pointer within index 0 of the buffer.
+        instr_valid_o = valid;
```
silabs-oysteink, line 280: Added to ensure we always signal valid for pointers, including misaligned ones.

Issue #676, opened 2022-09-28
Change 'parameter int' into 'parameter int unsigned'


Silabs-ArjanB:
This is to match the CORE-V-XIF specification. Also all other use of 'parameter int' does not make sense and should be changed into 'parameter int unsigned'.

This applies to the entire repos bhv, rtl, doc, etc. (and to surrounding modules instantiating this core)

Silabs-ArjanB:
Hi @silabs-oivind Can this issue be closed now?

silabs-oivind:
> Member

Yes.

Issue #39, opened 2021-04-28
Negative parameter indexes


MikeOpenHWGroup:
A recent addition to the CV32E40X (sha 4206fff4 on 2021-04-27) introduced "negative indexes" for the `PMA_CFG` parameter.  An example of this is `bhv/cv32e40x_wrapper.sv`:
```
module cv32e40x_wrapper
  import cv32e40x_pkg::*;
#(
  parameter NUM_MHPMCOUNTERS             =  1,
  parameter int unsigned PMA_NUM_REGIONS =  0,
  parameter pma_region_t PMA_CFG [PMA_NUM_REGIONS-1:0] = '{default:PMA_R_DEFAULT}
)
```
The default value of `PMA_NUM_REGIONS` is zero which means the range of `PMA_CFG` is [-1:0].  Most SV simulators will accept this, but Metrics DSIM interprets SV parameters as unsigned values (which, I believe is correct) and therefore the size of PMA_CFG is (2**31)-1 which exceeds a built-in DSIM limit.

Silabs-ArjanB:
Hi @silabs-oivind I believe my original proposal would prevent this.

silabs-oivind:
Hi @MikeOpenHWGroup,
I've updated the CV32E40X RTL to work around this.
However, if I read the section about range specification in the SystemVerilog LRM correctly, range boundaries are allowed to be any integer value, "positive, negative, or zero".

MikeOpenHWGroup:
Thanks @silabs-oivind, that did the trick.   Just for safety's sake it would also be good to do the same for the 40x_wrapper.

FWIW, I've looked into this and agree with you that the LRM supports negative range boundaries.  The Metrics DSIM simulator supports these as expected, so I was in error thinking this was the issue.   It turns out the problem is that DSIM interprets parameters as unsigned values, so if `PUM_NUM_REGIONS` is zero, then  `PMA_NUM_REGIONS-1` is resolved as a very large positive number, not -1.

PR #813, opened 2023-03-24
Clean up parameter types. Fix for issue #676

silabs-oivind:
Update most "parameter int" to "parameter int unsigned".
Binary parameters updated from "int" to "bit".
Note that PMA_NUM_REGIONS and DBG_NUM_TRIGGERS remain "parameter int" to avoid issues like the one reported in https://github.com/openhwgroup/cv32e40x/issues/39


sva/cv32e40x_mpu_sva.sv
```diff
@@ -24,13 +24,13 @@
 ////////////////////////////////////////////////////////////////////////////////
 
 module cv32e40x_mpu_sva import cv32e40x_pkg::*; import uvm_pkg::*;
-  #(  parameter int PMA_NUM_REGIONS              = 0,
+  #(  parameter int          PMA_NUM_REGIONS              = 0,
```
Silabs-ArjanB, line 27: IS_INSTR_SIDE should be 'bit' (might occur in other files as well)

Silabs-ArjanB (changes requested):

sva/cv32e40x_mpu_sva.sv
silabs-oivind: Updated.

PR #824, opened 2023-03-31
Changed tdata2 bits 4 and 6 to WARL for etrigger to enable trigger on…

silabs-oysteink:
… the new exception codes for misaligned atomics.


Silabs-ArjanB (changes requested):
Please add proper descriptions as well

Silabs-ArjanB (approved):

Issue #814, opened 2023-03-24
[XIF] Issue_req.id not updating and missing commit_valid


davidmallasen:
### Component
Issues in the RTL

### Steps to Reproduce
Git hash: b66765acf650493901b6d3e7b44eb37bd2598f4a

Hello,

In the same setup as issue #800 , the `id` of the instruction through the `issue_req` interface is not updating (is always 0) and therefore the second time an instruction is offloaded, it gets stuck waiting for a `commit_valid` that never arrives.

I'm running this snippet of code, where it gets stuck after the issue of the second load instruction:
![Screenshot at 2023-03-08 15-53-59](https://user-images.githubusercontent.com/15904722/223768748-7380c21c-2138-4a08-8743-826816f78afe.png)

Firstly there is a random offloading that gets killed instantly:
![cv32e40x_1](https://user-images.githubusercontent.com/15904722/227554300-0f30b80c-d14e-4e08-8024-9b036af4c278.png)

Then the first load instruction offloads and has to wait for the `mem_ready`:
![cv32e40x_2](https://user-images.githubusercontent.com/15904722/227554450-5e1b94b8-1f2f-4df5-b9d4-2fc4172b8448.png)

Then the `mem_ready` arrives, the instruction completes, and the second load is issued. However, the `commit_valid` for this second instruction never arrives because (I presume) the id is the same as the previous ones:
![cv32e40x_3](https://user-images.githubusercontent.com/15904722/227554662-c9f88ba3-dd4a-4231-b585-80fc183baeb5.png)

I attach a VDC dump of the execution.
[waveform.vcd.zip](https://github.com/openhwgroup/cv32e40x/files/11063526/waveform.vcd.zip)

Thanks!
David

Silabs-ArjanB:
Note that https://github.com/openhwgroup/cv32e40x/pull/825 was closed but its proposed fix has not been merged yet

davidmallasen:
Correct, the closing of the PR was an unintended side-effect, sorry about that @Silabs-ArjanB

PR #825, opened 2023-03-31
Fix #800 and #814

### LINKED ISSUE #800: Redundant mem_result when using CORE-V-XIF
### Component
Issues in the RTL

### Steps to Reproduce
Git hash: v0.8.0 974ca469f3a7598e99b5c6eb9d6a4b355fe8c9da

Hello,

I'm using CV32E40X inside [X-HEEP](https://github.com/esl-epfl/x-heep) together with a coprocessor that interacts with XIF. Through the XIF I'm receiving a redundant mem result from the CV32E40X that I believe should not be there. I'm not sure if it's a problem from my side or from the CV32E40X, but checking the waveforms I see everything correct from the coprocessor side (as I understand it).

I'm running the following snippet of code, where `plw` can be read as a normal lw but to an external register file in the coprocessor, and `peq` are equal operations between registers of this external RF that should write back a 0 or 1 to the GPRF of CV32E40X:
Screenshot at 2023-03-08 15-53-17

I atttach a screenshot of the waveform of the XIF signals:
Screenshot at 2023-03-08 15-50-11

The first 3 `issue_ready` - `issue_valid` handshake signals are for the 3 `plw`, and the next 3 are for the 3 `peq`. They are all signaled a `commit_kill = 0`, so the coprocessor initiates memory retrievals of the `plw` instructions with the 3 `mem_valid` - `mem_ready` handshakes. CV32E40X replies correctly through the `mem_result` interface, but after a while (right after the 3 `peq` instructions are executed and signaled via the `result` interface. There is an extra `mem_result_valid` where the yellow cursor is (identical to the last one) that should not be there, as there is no 4th mem request.

The output of the above program is: 
~~~
 PEQ test FAIL - Values: 7f939d17 7f6ec2bf 806e6c7b 1 1 1
~~~
Where the last 3 ones should be 0 1 1.

I also attach a VCD dump of this execution.

[waveform.vcd.zip](https://github.com/openhwgroup/cv32e40x/files/10922588/waveform.vcd.zip)

If it helps, this only happens with operations that end up writing back to the GPRF. Other snippets of code such as the following, work correctly.
Screenshot at 2023-03-08 15-53-59

Thanks!
David

davidmallasen:


rtl/cv32e40x_controller.sv
```diff
@@ -142,6 +142,8 @@ module cv32e40x_controller import cv32e40x_pkg::*;
 
   // eXtension interface
   if_xif.cpu_commit   xif_commit_if,
+  if_xif.cpu_mem       xif_mem_if,
```
silabs-oysteink, line 145: Connecting a if_xif.cpu_mem modport to multiple modules gives multiple drivers for the output signals for that modport. This could be broken up into individual signals, or a second interface instance with a  'monitor' modport could be instantiated at toplevel and connected to the 'cpu_mem' modport.

rtl/cv32e40x_controller_fsm.sv
```diff
@@ -1492,7 +1494,7 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
           commit_valid_q <= 1'b0;
           commit_kill_q  <= 1'b0;
         end else begin
-          if ((ex_valid_i && wb_ready_i) || ctrl_fsm_o.kill_ex) begin
+          if ((ex_valid_i && wb_ready_i) || ctrl_fsm_o.kill_ex || (xif_mem_if.mem_valid && xif_mem_if.mem_ready)) begin
```
silabs-oysteink, line 1497: The original intention of this code was to not clear the flags until the offloaded instruction (which is also tracked by the pipeline) leaves EX, or the EX stage is killed. With this change, the flag may be cleared early, even when WB is not ready to accept a new instruction.

rtl/cv32e40x_ex_stage.sv
```diff
@@ -428,6 +428,7 @@ module cv32e40x_ex_stage import cv32e40x_pkg::*;
       end else if (wb_ready_i) begin
         // we are ready for a new instruction, but there is none available,
         // so we introduce a bubble
+        ex_wb_pipe_o.xif_en      <= 1'b0;
```
silabs-oysteink, line 431: This should not be needed, instead all usage of ex_wb_pipe.xif_en inside the WB stage should be guarded by ex_wb_stage.instr.valid.

rtl/cv32e40x_id_stage.sv
```diff
@@ -683,6 +684,10 @@ module cv32e40x_id_stage import cv32e40x_pkg::*;
 
       end else if (ex_ready_i) begin
         id_ex_pipe_o.instr_valid            <= 1'b0;
+        id_ex_pipe_o.xif_en                 <= 1'b0;
+      end else if (xif_mem_if.mem_valid && xif_mem_if.mem_ready) begin
```
silabs-oysteink, line 688: At this point in the code, id_valid==0 and ex_ready == 0. This implies that the EX stage is not yet ready to accept a new instruction (id_ex_pipe_o must be kept stable), but the added code will insert a bubble into EX. What would happen if the offloaded load/store (in EX)  must stay in EX for multiple cycles while waiting for WB to become ready (load/store waiting for data_rvalid_i for instance)? Then the EX stage would be cleared by the inserted bubble, and the offloaded instruction would no longer be tracked by the pipeline.

silabs-oysteink: The LSU which drives the mem_ready signal is essential split between EX (address phase) and WB (response phase)

rtl/cv32e40x_id_stage.sv
```diff
@@ -683,6 +684,10 @@ module cv32e40x_id_stage import cv32e40x_pkg::*;
...
+        id_ex_pipe_o.instr_valid            <= 1'b0;
+        id_ex_pipe_o.xif_en                 <= 1'b0;
```
silabs-oysteink, line 690: Similar remark as for the controller, the id_ex_pipe.xif_en should be guarded by id_ex_pipe.instr_valid inside the EX stage instead.

silabs-oysteink (changes requested):

rtl/cv32e40x_controller_fsm.sv
davidmallasen: That `commit_valid_q` signal is negated in the `xif_commit_if.commit_valid` condition, so if you don't clear it in this case, you get a deadlock (see the red signals and the XIF signals):
commit_valid

rtl/cv32e40x_ex_stage.sv
davidmallasen: You are right, this was part of the debugging process and I forgot to remove it.

rtl/cv32e40x_id_stage.sv
davidmallasen: I'm not sure about that case in particular. However, without this the issue in #800 occurs. Maybe there is a better fix but we couldn't find it.

davidmallasen:
Thank you @silabs-oysteink . I replied to your comments and applied some changes. I cannot seem to find a barrier between these two issues (they must be linked) and this PR fixes them both.

davidmallasen:
To help with this debugging process, I set up an environment integrating FPNew and cv32e40x with the XIF using X-HEEP. I also have a test application that fails to execute properly. You can find the repo branch here: https://github.com/davidmallasen/F-HEEP/tree/bad-xif

I left some instructions and what to expect from the output here: https://github.com/davidmallasen/F-HEEP/issues/1

Issue #666, opened 2022-09-16
Check that none of the synchronous events that we check for in the FUNCTIONAL FSM state can be missed while in another state


Silabs-ArjanB:
See comment at bottom of https://github.com/openhwgroup/cv32e40x/pull/664.

PR #829, opened 2023-04-13
Fix for issue #666.

silabs-oysteink:
Added assertions to check that no synchronous events can be missed during any non-FUNCTIONAL FSM state.

### DIFF
--- a/sva/cv32e40x_controller_fsm_sva.sv
+++ b/sva/cv32e40x_controller_fsm_sva.sv
@@ -904,19 +904,23 @@ end
   assert property (@(posedge clk) disable iff (!rst_n)
                   (ctrl_fsm_cs == RESET)
                   |->
-                  !((pending_sync_debug && sync_debug_allowed) ||
-                    (exception_in_wb && exception_allowed)     ||
-                    (wfi_in_wb || wfe_in_wb)                   ||
-                    (fence_in_wb || fencei_in_wb)              ||
-                    (dret_in_wb)                               ||
-                    (csr_wr_in_wb_flush_i)                     ||
-                    (csr_flush_ack_q)                          ||
-                    (branch_taken_ex)                          ||
-                    (jump_taken_id)                            ||
-                    (clic_ptr_in_id || mret_ptr_in_id)         ||
-                    (mret_in_wb || mret_ptr_in_wb)             ||
-                    (clic_ptr_in_wb)                           ||
-                    (pending_single_step || etrigger_in_wb)))
+                  !((pending_sync_debug && sync_debug_allowed)                  ||
+                    (exception_in_wb && exception_allowed)                      ||
+                    (wfi_in_wb || wfe_in_wb)                                    ||
+                    (fence_in_wb || fencei_in_wb)                               ||
+                    (dret_in_wb)                                                ||
+                    (csr_wr_in_wb_flush_i)                                      ||
+                    (csr_flush_ack_q)                                           ||
+                    (branch_taken_ex)                                           ||
+                    (jump_taken_id)                                             ||
+                    (clic_ptr_in_id || mret_ptr_in_id)                          ||
+                    (mret_in_wb || mret_ptr_in_wb)                              ||
+                    (clic_ptr_in_wb)                                            ||
+                    (pending_single_step || etrigger_in_wb)                     ||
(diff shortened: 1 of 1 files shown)


sva/cv32e40x_controller_fsm_sva.sv
```diff
@@ -892,6 +899,136 @@ end
...
+                    (fence_in_wb || fencei_in_wb)              ||
+                    (dret_in_wb)                               ||
+                    (csr_wr_in_wb_flush_i)                     ||
+                    (csr_flush_ack_q)                          ||
+                    (branch_taken_ex)                          ||
+                    (jump_taken_id)                            ||
+                    (clic_ptr_in_id || mret_ptr_in_id)         ||
+                    (mret_in_wb || mret_ptr_in_wb)             ||
+                    (clic_ptr_in_wb)                           ||
+                    (pending_single_step || etrigger_in_wb)))
```
Silabs-ArjanB, line 919: Although the following synchronous events are implied by the ones you listed, I would still add them explcitly everywhere for easier review:

          if (clic_ptr_in_id && id_valid_i && ex_ready_i) begin
          if (mret_in_wb && !ctrl_fsm_o.kill_wb) begin
          if (mret_ptr_in_wb && !ctrl_fsm_o.kill_wb && !exception_in_wb) begin
          if(clic_ptr_in_wb && !ctrl_fsm_o.kill_wb && !exception_in_wb) begin

sva/cv32e40x_controller_fsm_sva.sv
```diff
@@ -892,6 +899,136 @@ end
...
+                    (mret_ptr_in_wb)                                                ||
+                    (clic_ptr_in_wb)                                                ||
+                    (pending_single_step && !pending_nmi && !(debug_cause_q == DBG_CAUSE_TRIGGER) && !(debug_cause_q == DBG_CAUSE_EBREAK)) // NMI can come at any time
+                    /*(etrigger_in_wb)*/)) // Left in on purpose, see comment above
+    else `uvm_error("controller", "Synchronous event during DEBUG_TAKEN(not single step)")
+
+
+    // Make sure no synchronous events are missed during DEBUG_TAKEN state caused by single stepping
+    // If the stepped instruction was a CSR write requiring pipeline flushing, the event
+    // csr_flush_ack_q will be active during DEBUG_TAKEN.
```
Silabs-ArjanB, line 1009: It is unclear why it is okay to ignore this event. Needs explanation and an assertion.

sva/cv32e40x_controller_fsm_sva.sv
```diff
@@ -892,6 +899,136 @@ end
...
+  //   - mret pointers in ID will be killed (unless the mret instruction itself finished in WB, then debug will not be allowed at all.)
+  //   - mret instructions in WB will be killed.
+
+  // The following events cannot be active during DEBUG_TAKEN:
+  //   - dret instructions cannot be in WB while in DEBUG_TAKEN (will be an illegal instruction exception instead)
+  //   - csr_flush_ack_q - Handshake flag one cycle after csr_wr_in_wb_flush_i.
+  //   - CLIC pointer in ID: When a CLIC pointer is in the pipeline, debug is not allowed
+  //   - mret or CLIC pointers in WB, debug is not allowed when these events are active.
+
+  // pending_single step may be true for the following conditions, but will not affect the debug cause:
```
Silabs-ArjanB, line 977: Can you explain why it will not effect debug cause and can be ignored? (I think the simple reason is that this is the lowest priority debug entry reason, right?)

Silabs-ArjanB (changes requested):

sva/cv32e40x_controller_fsm_sva.sv
silabs-oysteink: Updated all the new assertions with these events.

sva/cv32e40x_controller_fsm_sva.sv
silabs-oysteink: Added explanation

sva/cv32e40x_controller_fsm_sva.sv
silabs-oysteink: Added explanation (debug_cause_q was decided the cycle before entering DEBUG_TAKEN state)

Silabs-ArjanB (approved):

Issue #831, opened 2023-04-18
rvfi_intr.exception is kept high when an exception is followed by a trigger debug entry


silabs-mateilga:
![image](https://user-images.githubusercontent.com/59358563/232706993-04b3e760-f055-4da1-915c-c80df4b34eb7.png)

When an exception handler entry is immediately followed by a debug entry with cause trigger, in other words, when the first instruction in an exception handler is a debug trigger hit, rvfi_intr.exception remains high at the retire of the first instruction in the debug handler. The current expectation from verification asserts is that the rvfi_intr.exception signal should follow one retire after rvfi_trap.exception, unless an interrupt is signaled. The question here is if the verification expectation is wrong, or is the RTL implementation wrong?

Reproduce by formally verifying assert `...rvfi_i.rvfi_assert_i.a_sync_cause` with latest cv32e40s/dev branch of core-v-verif, at current date.

Silabs-ArjanB:
@silabs-mateilga Isn't this the scenario explained here?

![image](https://user-images.githubusercontent.com/40633348/232722584-c86f8f92-3526-44e0-81cf-6b08962869fc.png)

silabs-oysteink:
I think this issue is a duplicate of issue #355 on cv32e40s.
1: Core takes exception (RVFI starts tracking in_trap[])
2: First exception handler instruction gives a trigger match (IF/ID/EX is killed, wb_valid==1 to let RVFI know about the sync debug entry).

When rvfi_valid==1 for (2), rvfi_trap will indicate a debug entry while rvfi_intr will show the cause of exception from (1).  The issue arises if this first exception instruction is a POP for instance, which may have subops in multiple stages when the trigger match debug entry happens. This can cause the rvfi_intr information within RVFI to backpropagate to the IF stage, thus causing also the next rvfi_valid to carry the same rvfi_intr information.

Silabs-ArjanB:
@silabs-oysteink is there still an issue here then because 355 is marked as resolved? Is this on an older hash?

silabs-oysteink:
Issue 355 on cv3e40s was previously marked as resolved, but the label was removed as a new case came up that showed the same behaviour.

Silabs-ArjanB:
So, this is an RVFI issue (and not an RTL issue), right?

silabs-oysteink:
Yes, this is only for RVFI.

Silabs-ArjanB:
[...]

PR #835, opened 2023-04-19
For for issue #355 on cv32e40s.

silabs-oysteink:
Changed how RVFI backpropagates "in_trap" upon debug entry. Previously the backpropagation could cause already signaled "in_trap" to be backpropagated, causing multiple "rvfi_intr" for the same trap.

Should also fix issue #831


Silabs-ArjanB (approved):

Silabs-ArjanB:
Should be fixed by https://github.com/openhwgroup/cv32e40x/pull/835 (still needs to be merged to 40S)

Issue #766, opened 2023-01-23
CSR implementation differs from CV32E40S


Silabs-ArjanB:
Two issues really:

- The 40X CSR implementation differs from that of the 40S (use of MASK is not correct; i.e. non-implemented bits could be constant 1)
- The hardened vs. unhardened CSR implementations (on 40S) are not equivalent (hardened version is correct)

![image](https://user-images.githubusercontent.com/40633348/214018971-4ee26de3-7d62-4241-bc5d-973cb0b8d793.png)

silabs-oysteink:
PR #416 on CV32E40S addresses the issue of difference between hardened and unhardened CSRs.

For the masks, I think a mask bit of 1 should mean that there is an actual flop implemented for that bit, but it can still have a reset value of either 0 or 1. Further cleanup could introduce functions for use in _n value assignments (for both CV32E40S and CV32E40X), such that any non-zero reset values with mask bits cleared would never get set to zero (there is no case that I am aware of at the moment though).

`assign mtvec_n = csr_next_value(csr_wdata_int, CSR_MTVEC_MASK, MTVEC_RESET_VAL)` 
where the function would like something like

```
function automatic logic [31:0] csr_next_value
(
    logic [31:0] wdata,
    logic [31:0] mask,
    logic [31:0] reset_value
);
    return (wdata & mask) | (reset_value & ~mask);
endfunction
```

PR #836, opened 2023-04-20
Added usage of 'csr_next_value' function for almost all CSRs.

silabs-oysteink:
All CSR instances now use MASK and RESET_VAL. Some CSRs that are dependent on other CSR next values (mnxti, mscratchcsw[l]) do not use the function. Neither does the read-only CSRs that rely on inputs not coming from other CSRs (like mimpid, mip etc).

Should fix the remaining  fixes for issue #766.

SEC clean with CLIC==0 and CLIC==1.


silabs-oysteink:
I will revert some of these changes: I will only use the csr_next_value function wherever csr_wdata_int is used, and not for the cases where already resolved next_values or constants are used.

Silabs-ArjanB (approved):

PR #845, opened 2023-04-28
Fixed issue where rvfi_trap.debug would not be set for an etrigger.

silabs-oysteink:
Added assertions for the specific issue, and updated assertions to accomodate the change.

### DIFF
--- a/bhv/cv32e40x_rvfi.sv
+++ b/bhv/cv32e40x_rvfi.sv
@@ -124,6 +124,7 @@ module cv32e40x_rvfi
    input logic                                nmi_is_store_i,         // regular NMI type
    input logic                                debug_mode_q_i,
    input logic [2:0]                          debug_cause_n_i,
+   input logic                                etrigger_in_wb_i,
 
    // Interrupt Controller probes
    input logic [31:0]                         irq_i,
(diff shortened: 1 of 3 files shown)


bhv/cv32e40x_rvfi.sv
```diff
@@ -872,13 +873,30 @@ module cv32e40x_rvfi
...
       rvfi_trap_next.debug       = 1'b1;
       rvfi_trap_next.debug_cause = DBG_CAUSE_STEP;
 
       // In the case of an exception in WB and pending single step, both the exception and the debug flag will be set
     end
 
+
+    // Check for etrigger debug entry, need to include the actual debug_cause_n, as etrigger has lower priority
+    // to enter debug than for instance external debug request.
+    if((etrigger_in_wb_i && single_step_allowed_i) && (debug_cause_n_i == DBG_CAUSE_TRIGGER)) begin
```
Silabs-ArjanB, line 889: Why is there this '&& single_step_allowed_i'? That seems odd as qualification for etrigger?

Silabs-ArjanB (commented):
See question

bhv/cv32e40x_rvfi.sv
silabs-oysteink: It is used in the controller_fsm for gating both single step and etrigger. I think it would make most sense to have an extra 'etrigger_allowed' inside the controller_fsm, or rename the current to someting like 'etrigger_and_single_step_allowed'. The signal in question is tied to 1'b1 inside the controller.

Silabs-ArjanB (approved):

PR #851, opened 2023-05-09
Add assertions for data bus vs RVFI consistency

silabs-oivind:


silabs-oysteink (approved):

sva/cv32e40x_rvfi_sva.sv
```diff
@@ -426,6 +433,315 @@ end
...
+                                                          (rvfi_trap.debug_cause == 3'h2))));  // Debug trigger match
+
+    assign rvfi_mem.addr  = rvfi_mem_addr;
+    assign rvfi_mem.rmask = rvfi_mem_rmask;
+    assign rvfi_mem.wmask = rvfi_mem_wmask;
+    assign rvfi_mem.rdata = rvfi_mem_rdata;
+    assign rvfi_mem.wdata = rvfi_mem_wdata;
+    assign rvfi_mem.prot  = rvfi_mem_prot;
+
+`ifdef RVFI_SVA_CONST_DATA_OBI_GNT
```
Silabs-ArjanB, line 520: Can this ifdef be avoided somehow; what are the alternatives we have? We worked hard to get rid of all such non-standard ifdefs that are not documented anywhere. Currently the only ifdef used in the entire 40x repos is on RISCV_FORMAL (and even that is non-standard)

bhv/cv32e40x_rvfi.sv
```diff
@@ -1275,11 +1275,12 @@ module cv32e40x_rvfi
...
-          rvfi_mem_rmask[ (4*(memop_cnt+1))-1 -:  4] <= mem_access_blocked_wb ? '0 : mem_rmask [STAGE_WB];
-          rvfi_mem_wmask[ (4*(memop_cnt+1))-1 -:  4] <= mem_access_blocked_wb ? '0 : mem_wmask [STAGE_WB];
-          rvfi_mem_addr [(32*(memop_cnt+1))-1 -: 32] <= ex_mem_trans.addr;
-          rvfi_mem_wdata[(32*(memop_cnt+1))-1 -: 32] <= ex_mem_trans.wdata;
-          rvfi_mem_prot [ (3*(memop_cnt+1))-1 -:  3] <= ex_mem_trans.prot;
+          rvfi_mem_rmask[ (4*(memop_cnt+1))-1 -:  4]   <= mem_access_blocked_wb ? '0 : mem_rmask [STAGE_WB];
+          rvfi_mem_wmask[ (4*(memop_cnt+1))-1 -:  4]   <= mem_access_blocked_wb ? '0 : mem_wmask [STAGE_WB];
+          rvfi_mem_addr [(32*(memop_cnt+1))-1 -: 32]   <= ex_mem_trans.addr;
+          rvfi_mem_wdata[(32*(memop_cnt+1))-1 -: 32]   <= ex_mem_trans.wdata;
+          // Using (2*memop_cnt+memop_cnt) rather than 3*memop_cnt. This is a workaround to avoid blackboxed multiplier in the slice boundary calculations
```
Silabs-ArjanB, line 1282: How does the multiplier result propagate into this (into what signal does it propagate)? The multiplier might propagate in some OBI signals but it will of course not affect actual transfers, so could an alternative be to only check actual transfers?

silabs-oivind: We are only checking actual transfers. The multiplication 3*memop_cnt could be blackboxed in formal, resulting in wrong slice boundaries, so e.g. rvfi_mem_prot[-1:-3] is updated instead of rvfi_mem_prot[2:0].

Silabs-ArjanB: I misunderstood which multiplier you were talking about. Quite sad that the tool black boxes 3 * ...

sva/cv32e40x_rvfi_sva.sv
silabs-oivind: I've updated as we discussed, adding a qualification signal on the write asserts. Removed ifdef and assume on obi.gnt

Silabs-ArjanB (approved):

Issue #808, opened 2023-03-18
Vivado simulation crashes due to Verilog arrays defined in LSB format in cv32e40x_alignment_buffer.sv


Srrbli-23:
### Component

**Component:RTL**
rtl/cv32e40x_alignment_buffer.sv

### Steps to Reproduce
Git hash: v0.8.0 `b22444c7f517964313286af485e4a4ddb62467cc` [https://github.com/openhwgroup/cv32e40x/commit/b22444c7f517964313286af485e4a4ddb62467cc](url)

Hello,

As a beginner testing the cv32e40x project on Xilinx's FPGA, I found that the project includes the file cv32e40x_alignment_buffer.sv, which contains two arrays defined in LSB format. These arrays cause Vivado to crash during simulation.

```systemverilog
  inst_resp_t [0:ALBUF_DEPTH-1]  resp_q;
  logic [0:ALBUF_DEPTH-1]        valid_n,   valid_int,   valid_q;
```

When running the simulation in Vivado, the following error message will be prompted, and the simulation will crash:
```
FATAL_ERROR: Vivado Simulator kernel has discovered an exceptional condition from which it cannot recover. Process will terminate. For technical support on this issue, please open a WebCase with this project attached at http://www.xilinx.com/support.
Time: 0 ps  Iteration: 0  Process: /path/to/u0/if_stage_i/prefetch_unit_i/alignment_buffer_i/NetRegassign166_292
  File: /path/to/cv32e40x_alignment_buffer.sv
```

![reproduce](https://user-images.githubusercontent.com/29352211/226118724-98170c17-8a2a-4888-a6bf-73d29f67af28.png)

To resolve this issue, it is recommended that the arrays be defined in MSB format, for example:
```
  inst_resp_t [ALBUF_DEPTH-1:0]  resp_q; 
  logic [ALBUF_DEPTH-1:0]        valid_n,   valid_int,   valid_q;
```

![fix](https://user-images.githubusercontent.com/29352211/226118733-0590c4dc-15f2-4f24-9242-526fde12ac9c.png)

I believe this may be a Vivado simulation software issue. However, updating the arrays to be defined in MSB format can facilitate debugging on an FPGA.

Thanks!
Srrb

Silabs-ArjanB:
Vivado issue should be worked around with https://github.com/openhwgroup/cv32e40x/pull/864

PR #864, opened 2023-05-30
Fix for issue #808.

silabs-oysteink:
Changed two LSB format vectors to MSB format in alignment buffer.

SEC clean.


silabs-oysteink:
Added do-not-merge label as issues were detected in formal.

Silabs-ArjanB (approved):

PR #885, opened 2023-07-05
Updates to yaml.m4

silabs-oysteink:
Added enums for exception codes and mpp.
Renamed RESERVED_XX_YY to RESVERVED<N>.
Added parameter to allow a different view of mtvec and mtvt for verification and non-verification headers.

### DIFF
--- a/yaml/csr.yaml.m4
+++ b/yaml/csr.yaml.m4
@@ -3067,14 +3067,14 @@ ifelse(eval(CLINT != 0), 1, [[[
   address: 0x305
   privilege_mode: M
   rv32:
-    - field_name: BASE31TO7
+    - field_name: BASE_31_7
       description: >
         Trap-handler base address, always aligned to 128 bytes
       type: WARL
       reset_val: 0  # Note: assumes mtvec_i == 0
       msb: 31
       lsb: 7
-    - field_name: BASE6TO2
+    - field_name: BASE_6_2
       description: >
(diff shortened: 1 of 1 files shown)


silabs-oivind (approved):

yaml/csr.yaml.m4
```diff
@@ -3096,6 +3101,7 @@ ifelse(eval(CLIC != 0), 1, [[[
   address: 0x305
   privilege_mode: M
   rv32:
+ifelse(eval(VERIF_HEADER != 0), 1, [[[
```
Silabs-ArjanB, line 3104: Can't this ifdef be prevented by using another legalize function on bits 31-6 combined?

yaml/csr.yaml.m4
```diff
@@ -3139,6 +3154,7 @@ ifelse(eval(CLIC != 0), 1, [[[
   address: 0x307
   privilege_mode: M
   rv32:
+ifelse(eval(VERIF_HEADER != 0), 1, [[[
```
Silabs-ArjanB, line 3157: Same remark here; is an ifselse really needed?

yaml/csr.yaml.m4
silabs-oysteink: The idea is that the verification headers may benefit from having the base bitfield split in two, while the sw headers are likely better off with one base field.

Silabs-ArjanB (approved):

Issue #892, opened 2023-07-21
Explain RVFI corner case better


Silabs-ArjanB:
PR https://github.com/openhwgroup/cv32e40x/pull/890 in itself looks good to me, but I don't understand the comments explaining that  ctrl_fsm_i.debug_cause will not capture this corner case.

Can further comments be added (maybe after running formal to find a good counter example) on why this code:

rvfi_trap_next.debug_cause = (ebreak_in_wb_i && debug_mode_q_i) ? DBG_CAUSE_EBREAK : ctrl_fsm_i.debug_cause;

debug_cause[STAGE_IF] <= (ebreak_in_wb_i && debug_mode_q_i) ? DBG_CAUSE_EBREAK : ctrl_fsm_i.debug_cause;

cannot simply be:

rvfi_trap_next.debug_cause = ctrl_fsm_i.debug_cause;

debug_cause[STAGE_IF] <= ctrl_fsm_i.debug_cause;

PR #919, opened 2023-08-24
Fix for issue #892

silabs-oysteink:
Changed from:
(ebreak_in_wb_i && debug_mode_q_i) ? DBG_CAUSE_EBREAK : ctrl_fsm_i.debug_cause

to:
ctrl_fsm_i.debug_cause

in both lines in question. Checked with a temporary assertion to check for correctness.


Silabs-ArjanB (changes requested):
Fix comments as well

Silabs-ArjanB (approved):

PR #925, opened 2023-08-31
Removed CSR mscratchcsw.

silabs-oysteink:


yaml/csr.yaml.m4
```diff
@@ -4487,22 +4487,6 @@ ifelse(eval(CLIC != 0), 1, [[[
       lsb: 0
 ]]])
```
Silabs-ArjanB, line 4488: Not okay to remove this from the yaml file as we try to have the same yaml between 40x and 40s. The CSR should be excluded for non-CLIC and if user mode is not supported.

Only this file should get merged to the 40S, not the other files

rtl/cv32e40x_i_decoder.sv
```diff
@@ -379,8 +379,8 @@ module cv32e40x_i_decoder import cv32e40x_pkg::*;
           endcase
 
           if (CLIC) begin
-            // The mscratchcsw[l] CSRs are only accessible using CSRRW with neither rd nor rs1 set to x0
-            if ((instr_rdata_i[31:20] == CSR_MSCRATCHCSW) || (instr_rdata_i[31:20] == CSR_MSCRATCHCSWL)) begin
+            // The mscratchcswl CSR is only accessible using CSRRW with neither rd nor rs1 set to x0
+            if ((instr_rdata_i[31:20] == CSR_MSCRATCHCSWL)) begin
```
Silabs-ArjanB, line 383: Remove double brackets

Silabs-ArjanB (changes requested):

yaml/csr.yaml.m4
silabs-oysteink: Added mscratchcsw back in and guarded with CLIC and UMODE.

rtl/cv32e40x_i_decoder.sv
silabs-oysteink: Fixed

Silabs-ArjanB (approved):

PR #928, opened 2023-08-31
Updates for CLIC spec (august 23)

silabs-oysteink:
mcause.minhv only set when taking an exception for a CLIC pointer or mret pointer restart. When jumping due to an mret, mcause.minhv is checked regardless of mcause.mpp.

### DIFF
--- a/rtl/cv32e40x_controller_fsm.sv
+++ b/rtl/cv32e40x_controller_fsm.sv
@@ -75,7 +75,6 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
   input  align_status_e align_status_wb_i,          // Aligned status (atomics) in WB
   input  logic [31:0]   wpt_match_wb_i,             // LSU watchpoint trigger (WB)
 
-
   // From LSU (WB)
   input  logic        data_stall_wb_i,            // WB stalled by LSU
   input  logic        lsu_valid_wb_i,             // LSU instruction in WB is valid
@@ -282,8 +281,6 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
   // index 0 for non-vectored CLINT mode and CLIC mode, 0xF for vectored CLINT mode
   assign ctrl_fsm_o.nmi_mtvec_index = (mtvec_mode_i == 2'b01) ? 5'hF : 5'h0;
 
-
-
   ////////////////////////////////////////////////////////////////////
   // ID stage
 
@@ -291,8 +288,6 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
   // Checking validity of jump/mret instruction with if_id_pipe_i.instr_valid and the respective alu_en/sys_en.
   // Using the ID stage local instr_valid would bring halt_id and kill_id into the equation
   // causing a path from data_rvalid to instr_addr_o/instr_req_o/instr_memtype_o via pc_set.
-
-
 
   assign sys_mret_id = sys_en_id_i && sys_mret_id_i && if_id_pipe_i.instr_valid;
   assign jmp_id      = alu_en_id_i && alu_jmp_id_i  && if_id_pipe_i.instr_valid;
@@ -429,7 +424,7 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
   // This CSR bit shall not be gated by debug mode or step without stepie
   assign ctrl_fsm_o.pending_nmi = nmi_pending_q;
 
-  // Debug //
+  // Debug
 
   // Single step will need to finish insn in WB, including LSU
   // LSU will now set valid_1_o only for second part of misaligned instructions.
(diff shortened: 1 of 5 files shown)


rtl/cv32e40x_controller_fsm.sv
```diff
@@ -864,8 +861,8 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
             ctrl_fsm_o.csr_save_cause = !debug_mode_q; // Do not update CSRs if in debug mode
             ctrl_fsm_o.csr_cause.exception_code = exception_cause_wb;
 
-            // Keep mcause.minhv when taking exceptions and interrupts, only cleared on successful pointer fetches or CSR writes.
-            ctrl_fsm_o.csr_cause.minhv  = mcause_i.minhv;
+            // Set mcause.minhv if exception is for a CLIC or mret pointer. Otherwise clear it
+            ctrl_fsm_o.csr_cause.minhv = (ex_wb_pipe_i.instr_meta.clic_ptr || ex_wb_pipe_i.instr_meta.mret_ptr);
```
Silabs-ArjanB, line 865: Why using the unqualified signals here?

rtl/cv32e40x_controller_fsm.sv
```diff
@@ -1465,6 +1456,9 @@ module cv32e40x_controller_fsm import cv32e40x_pkg::*;
   assign ctrl_fsm_o.debug_running   = debug_fsm_cs[RUNNING_INDEX];
   assign ctrl_fsm_o.debug_halted    = debug_fsm_cs[HALTED_INDEX];
 
+  // Tie off unused signals
+  assign unused_signals = clic_ptr_in_wb;
```
Silabs-ArjanB, line 1460: I think this signal should actually be used

sva/cv32e40x_cs_registers_sva.sv
```diff
@@ -163,6 +163,25 @@ module cv32e40x_cs_registers_sva
...
+    // Whenever mepc is written by HW, mcause.minhv must also be written with the correct value.
+    // mcause.minhv is only set when an exception is taken on a CLIC or mret pointer
+    // SW can still write both mepc and mcause independently
+    property p_mepc_minhv_write;
+      @(posedge clk) disable iff (!rst_n)
+      (
+        mepc_we   // mepc is written
+        |->
+        (mcause_we && // mcause is written
+        mcause_n.minhv == ((ex_wb_pipe_i.instr_meta.clic_ptr || ex_wb_pipe_i.instr_meta.mret_ptr) && // pointer in WB
```
Silabs-ArjanB, line 175: Use qualified signals

sva/cv32e40x_cs_registers_sva.sv
```diff
@@ -163,6 +163,25 @@ module cv32e40x_cs_registers_sva
     a_mcause_mstatus_alias: assert property(p_mcause_mstatus_alias)
       else `uvm_error("cs_registers", "mcause.mpp and mcause.mpie not aliased correctly")
 
+    // Whenever mepc is written by HW, mcause.minhv must also be written with the correct value.
```
Silabs-ArjanB, line 166: Prove other way around as well (if minhv is written mepc should be written)

Silabs-ArjanB (changes requested):

rtl/cv32e40x_controller_fsm.sv
silabs-oysteink: 'exception_in_wb' on line 846 is already qualified with ex_wb_pipe_i.instr_valid. It shouldn't be needed to also factor it into line 865.

rtl/cv32e40x_controller_fsm.sv
silabs-oysteink: Agreed, it could be used for detecting 'clic_ptr_in_pipeline'. I will refactor code to use this signal.

rtl/cv32e40x_controller_fsm.sv
Silabs-ArjanB: It is not strictly needed indeed, but it would be more readable

silabs-oysteink: Updated code

rtl/cv32e40x_controller_fsm.sv
silabs-oysteink: Removed the 'unused_signal'

sva/cv32e40x_cs_registers_sva.sv
silabs-oysteink: Updated

sva/cv32e40x_cs_registers_sva.sv
silabs-oysteink: Added assert to prove that whenever mcause.minhv changes mepc must also be written.

PR #932, opened 2023-09-05
Remove mscratchcsw register from manual

silabs-halfdan:
Remove mscratchcsw register from manual as it is meant for priv mode change. This has already been done in rtl.


Silabs-ArjanB (changes requested):
This PR only partially remove this CSR from the documentation.

Instead all occurences should be looked at and the part you removed should be moved into the 'only:: USER' section instead

PR #946, opened 2023-09-13
Misc sva updates

silabs-oivind:


silabs-oysteink (approved):

sva/cv32e40x_controller_fsm_sva.sv
```diff
@@ -808,14 +803,13 @@ end
                     (sequence_in_progress_wb |-> !ctrl_fsm_o.kill_wb))
     else `uvm_error("controller", "WB killed while sequence was in progress")
 
-/* todo: Bring back in if id_stage_haltable_alt can be correctly calculated.
   // Check that we do not allow ID stage to be halted for pending interrupts/debug if a sequence is not done
   // in the ID stage.
   a_id_stage_haltable:
   assert property (@(posedge clk) disable iff (!rst_n)
-                    (id_stage_haltable_alt == id_stage_haltable))
+                    (id_stage_haltable |-> id_stage_haltable_alt))
```
Silabs-ArjanB, line 810: How about the other way around?

Silabs-ArjanB (commented):
See question in review

sva/cv32e40x_controller_fsm_sva.sv
silabs-oivind: The reasoning for changing to a one-way assertion is that we want to keep id_stage_haltable_alt as disconnected from the pipeline as possible, this to prevent it from just being a copy of the RTL version.
To check both ways, id_stage_haltable_alt would have to take e.g. pipeline handshakes into account, which would result in it being more similar to the RTL implementation.
The most important thing is still being checked; that the design isn't halting ID when it's not allowed to.

Silabs-ArjanB (approved):

PR #949, opened 2023-09-15
Add cv32e40x_parameter_sva.sv

silabs-oivind:
Part of todo fixing on e40s
a_pma_valid_config was removed since breaking this will result in compile error.


Silabs-ArjanB (approved):
Hi @silabs-hfegran Are you okay with the use of $fatal in this file?

silabs-hfegran:
> Hi @silabs-hfegran Are you okay with the use of $fatal in this file?

For now I think this is OK.

PR #957, opened 2023-09-29
CSR Yaml Updates

silabs-robin:
This PR updates `csr.yaml.m4`.

`mcause.MPP` had U-mode as reset value, on the X.
A `warl_legalize` produced wrong generated test code on a WRLR field.
Special CLIC access behavior also generated wrong test code.

### DIFF
--- a/yaml/csr.yaml.m4
+++ b/yaml/csr.yaml.m4
@@ -4139,7 +4139,7 @@ ifelse(eval(CLIC != 0), 1, [[[
       description: >
           Alias for mstatus.MPP
       type: WARL
-      reset_val: 0
+      reset_val: 3
       msb: 29
       lsb: 28
       warl_legalize: |
@@ -4183,7 +4183,11 @@ ifelse(eval(UMODE != 0), 1, [[[
     - field_name: Exccode
       description: >
           Exception Code
+ifelse(eval(VERIF_HEADER != 0), 1, [[[
+      type: WARL  # "warl_legalize" does nothing without "WARL".
+]]], [[[dnl
       type: WLRL
+]]])
       reset_val: 0
       msb: 11
       lsb: 0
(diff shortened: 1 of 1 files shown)


yaml/csr.yaml.m4
```diff
@@ -4139,7 +4139,7 @@ ifelse(eval(CLIC != 0), 1, [[[
       description: >
           Alias for mstatus.MPP
       type: WARL
-      reset_val: 0
+      reset_val: 3
```
Silabs-ArjanB, line 4142: Make depedend on param so that it can be merged to 40S

silabs-robin, line 4142: I'll wrap this S/X-difference into a parameter.

yaml/csr.yaml.m4
```diff
@@ -4410,7 +4414,7 @@ ifelse(eval(CLIC != 0), 1, [[[
         val_out = 0
 ]]])
 
-ifelse(eval(CLIC != 0), 1, [[[
+ifelse(eval(CLIC != 0 && READONLY != 0), 1, [[[
```
Silabs-ArjanB, line 4417: How/when/why is READONLY set/cleared?

Silabs-ArjanB (changes requested):

yaml/csr.yaml.m4
silabs-robin: > [How] is READONLY set/cleared?

`m4 .... -D READONLY=0`  (Or 1.)

> [When] is READONLY set/cleared?

Anyone who uses m4 to generate a specific YAML **must** define all variables which are used in `eval` expressions.

> [Why] is READONLY set/cleared?

If it is not defined one gets an error `[...] bad expression in eval`.
We **clear** it in verif when generating the CSR access test, because the test generator doesn't seem to support all the variation found in RISC-V CSR behavior.

yaml/csr.yaml.m4
silabs-robin: The `mcause` reset value is the same for both cores?

yaml/csr.yaml.m4
silabs-robin: It is not new.

yaml/csr.yaml.m4
silabs-robin: No difference between X and S.

Silabs-ArjanB (approved):

PR #958, opened 2023-10-02
Introduced type for lsu_err_wb

silabs-oysteink:
SEC clean


silabs-robin (commented):
Does hello world pass and is formal prop still functional?

silabs-oysteink:
ci_check passes and formal prop seems good to me.

PR #977, opened 2023-11-22
BUGFIX: for cm.mva01s when rs1 == rs2

silabs-oysteink:
Both cm.mva01s and cm.mvsa01 flagged illegal instruction if rs1 == rs2.

With this fix cm.mva01s allows rs1 == rs2, and the core will not flag illegal instruction.

### DIFF
--- a/rtl/cv32e40x_sequencer.sv
+++ b/rtl/cv32e40x_sequencer.sv
@@ -83,7 +83,8 @@ module cv32e40x_sequencer import cv32e40x_pkg::*;
 
   logic                ready_fsm;
 
-  logic                dmove_legal_dest;
+  logic                dmove_legal_dest_s2a;
+  logic                dmove_legal_dest_a2s;
   logic                pushpop_legal_rlist;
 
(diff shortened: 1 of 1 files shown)


rtl/cv32e40x_sequencer.sv
```diff
@@ -136,11 +137,14 @@ module cv32e40x_sequencer import cv32e40x_pkg::*;
   assign pushpop_legal_rlist = (RV32 == RV32I) ? (rlist > 4'h3) :
                                (rlist < 4'h7) && (rlist > 4'h3);
 
-  // rs1 must be different from rs2
+  // cm.mva01s: allows rs1' and rs2' to be the same register
```
silabs-hfegran, line 140: Zc spec names these registers `r1s'` and `r2s'` as they have a different representation from what `rs1'` and `rs2'` normally represents. Only referring to the comment, the code itself looks good to me.

silabs-hfegran (approved):

silabs-oivind (approved):

Silabs-ArjanB (approved):

PR #982, opened 2024-08-08
Tracer update

silabs-oivind:
Update string casting in cv32e40x_rvfi_sim_trace for better simulator compatibility.
Moving trace_t typedef to package to enable reuse outside of cv32e40x_rvfi_sim_trace

### DIFF
--- a/bhv/cv32e40x_rvfi_sim_trace.sv
+++ b/bhv/cv32e40x_rvfi_sim_trace.sv
@@ -48,15 +48,6 @@ module cv32e40x_rvfi_sim_trace
    input logic [32*NMEM-1:0] rvfi_mem_wdata
    );
 
-  typedef struct      {
-    bit [32*16-1:0]  c_file;
-    bit [32*16-1:0]  c_func;
-    int              c_line;
-    logic [31:0]     addr;
-    logic [31:0]     mcode;
-    bit [32*16-1:0]  asm;
-  } trace_t;
-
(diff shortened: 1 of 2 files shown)


silabs-oysteink (approved):

bhv/cv32e40x_rvfi_sim_trace.sv
```diff
@@ -182,16 +173,21 @@ module cv32e40x_rvfi_sim_trace
...
                   for (int i=0; i < asmlen; i++) begin
                     if(asms[i] == "#") break; // Disregard comments
 
                     // Concatenate assembly instruction
-                    $cast(tmp_trace.asm,  $sformatf("%-0s %-0s", tmp_trace.asm, asms[i]));
+                    tmp_asm_string = $sformatf("%-0s %-0s", tmp_asm_string, asms[i]);
                   end
 
+                  // Cast string using streaming operator
+                  tmp_trace.asm = { >> {tmp_asm_string}};
```
silabs-hfegran, line 188: Is this safe? No type compatibility check is performed as the raw bit-values are streamed into a different type.

silabs-oivind: No:) But in this context I believe it's OK.
The reason we're casting from string to bit array is that some waveform viewers do not display strings in a good way.
We're casting from a string to a bit array that will be displayed as a string in waveform viewers. 
The type check in $cast fails in some simulators, so the point here is actually to relax (skip) type checking.

PR #987, opened 2024-10-30
Improved ebreak behavior description

Silabs-ArjanB:


docs/user_manual/source/debug.rst
```diff
@@ -185,5 +185,5 @@ Scenario 3 : Exit Program Buffer & Restart Debug Code
 
 Executing the ``ebreak`` instruction when the core is in Debug Mode shall result in the following actions:
 
-- The core remains in Debug Mode and execution jumps back to the beginning of the debug code located at ``dm_halt_addr_i``
+- The core re-enters Debug Mode (without actually having ever left Debug Mode) by jumping back to the beginning of the debug code located at ``dm_halt_addr_i``
```
silabs-robin, line 188: This new phrasing is better IMO.

Good clarification about not leaving debug mode too.

silabs-robin:
Context (for posterity):

"For the first instruction after entering debug, the rvfi_dbg signal contains the debug cause (see table below). The signal is otherwise 0." - UM

"an ebreak in debug mode will be reported via rvfi_dbg" - UM

^ The ebreak can only be reported on `rvfi_dbg` after _entering_ debug mode. Hence this PR's rephrasing.

PR #988, opened 2024-11-05
Added dedicated LSU operand registers

Seek64:
Based on a discussion with @Silabs-ArjanB 

This PR introduces dedicated registers for LSU operands, which prevents the operands of non-LSU instructions from leaking to the data memory interface. 
In the current version, internal data is always encoded as an address and forwarded, even if _data_req_o_ is low.
A connected device that does not conform to the OBI protocol (because it's buggy or malicious) could use this covert channel to undermine security features such as _data_ind_timing_.

I have implemented the identical fix in my setup for cv32e40s to ensure that non-LSU operands do not leak to the data memory interface. 
There may still be invalid memory requests and thus stale data, but only for LSU instructions.
Also, I have not yet considered compressed instructions.

I have not run any functional verification with this fix implemented.

### DIFF
--- a/rtl/cv32e40x_id_stage.sv
+++ b/rtl/cv32e40x_id_stage.sv
@@ -516,6 +516,9 @@ module cv32e40x_id_stage import cv32e40x_pkg::*;
       id_ex_pipe_o.csr_en                 <= 1'b0;
       id_ex_pipe_o.csr_op                 <= CSR_OP_READ;
 
+      id_ex_pipe_o.lsu_operand_a          <= 32'b0;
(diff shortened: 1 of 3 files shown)


Silabs-ArjanB:
Hi @Seek64 Thank you for your pull request. At first glance it looks at least functionally correct (but we will do a formal run to prove that). Also at first glance I think it must be possible to sometimes prevent updating the ALU operands when load/store instructions are being execute (to save power). That additional power savings is of course unrelated to the issue that you were trying to solve, so maybe we can just to a separate pull request for that later. It will take some time before we can study this further and do the merge.

PR #992, opened 2026-03-18
[XIF] Making the XIF ID unique

Joao-Pedro-Cabral:
As stated on the file `cv32e40x_if_stage.sv` on the commit `52c86ae0c95cadd005fae3f8c7425a72d0687d0c`:
_// TODO:XIF assert that the oustanding IDs are unique_

And in my personal setup I got situations where the outstanding IDs are not unique.
This behavior can happens if in one cycle the ID offloads a instruction via XIF, then `xif_offloading_id_i` turns on.
However, it is possible for the ID stage updates its content to the next instruction, while the IF is stalled. In this situation,  `xif_offloading_id_i` will be turned off before the IF stage ready to increment the `xif_id` register.

My solution is very simple, basically buffering the `xif_offloading_id_i` signal if the IF stage is not ready, and clearing it when IF is ready.


Joao-Pedro-Cabral:
Sorry, can you help me with the ECA, I've never used this type of tool :/ ?
I created the account, linked my github account and accepted the ECA, but it's not working yet.

